PR #25, opened 2025-04-04
Update navigation bar and footer to support instance customization

tamthaihoangminh:
Introduce instance customization support for navigation bar and footer components to align with playground's extension features.

Part of resolving issue #18


tuan-hda (approved):
LGTM

PR #59, opened 2025-04-23
Fix #58 : useListAllModels hook to only query when access is truthy

### LINKED ISSUE #58: Fix: useListAllModels hooks use wrong condition filter: "enabled: access !== undefined"

tamthaihoangminh:
This close #58 

This PR fixes a bug where the useListAllModels hook was using enabled: access !== undefined, which allowed the query to run even when access was null or false. The condition is now updated to "queryKey: ['listAllModels', self?.id],", so the query only runs when the user access change.


tuan-hda:
LGTM

PR #175, opened 2025-06-30
Replace Apache header with MIT and add LICENSE file

tri2510:
Replace Apache license header with MIT header and add MIT headers to all source files


tamthaihoangminh (commented):
Just adding MIT license header. LGTM

PR #335, opened 2026-02-27
Refactor template tabs and fix overlay issue

trantienloi2404:
- Refactor model/prototype tabs layout from grid to flex for better responsiveness.
- Fix overlay issue after closing modal
- Fix create model by using template issue

### DIFF
--- a/frontend/src/components/atoms/dialog.tsx
+++ b/frontend/src/components/atoms/dialog.tsx
@@ -52,16 +52,6 @@ const DialogContent = React.forwardRef<
         className,
       )}
       {...props}
-      // Radix DismissableLayer sets `pointer-events: none` on <body> when modal dialogs open.
-      // Multiple Radix components (dialog, dropdown-menu, select) each bundle their own
-      // DismissableLayer copy with separate state, causing stale `pointer-events: none` to
-      // persist after close (e.g. opening this dialog from a DropdownMenu).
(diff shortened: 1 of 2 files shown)


NhanLuongBGSV:
## Follow-up: Replaced global CSS workaround with proper fix (6c73090)

The `body { pointer-events: auto !important; }` in `index.css` has been **removed** and replaced with a proper component-level fix.

### What changed

**`frontend/src/components/atoms/dialog.tsx`** — Added a cleanup `useEffect` in `DialogContent` that restores `document.body.style.pointerEvents` when the component unmounts while body is still stuck with `pointer-events: none`:

```tsx
React.useEffect(() => {
  return () => {
    if (document.body.style.pointerEvents === 'none') {
      document.body.style.pointerEvents = ''
    }
  }
}, [])
```

### Why this is better than the global CSS override

- **Targeted** — Only runs when a dialog actually unmounts, not a blanket `!important` override
- **Covers all dialogs** — `DialogContent` is the shared base component used by both `DaDialog` (wrapper) and all 12 files that use raw `Dialog` directly
- **No side effects** — Won't interfere with any legitimate `pointer-events: none` usage on body outside of dialog contexts
- **Self-documenting** — The fix lives right next to the Radix integration code where the bug originates

**`frontend/src/index.css`** — Removed the global `body { pointer-events: auto !important; }` workaround entirely.

PR #337, opened 2026-03-03
Fix export prototype in the multiple files case

trantienloi2404:
- Add logic to handle export/import prototype in the multiple files case


NhanLuongBGSV:
@trantienloi2404 Hi Loi, please make small improvements as suggested above

Issue #50, opened 2025-04-21
Bug: After creating new instance, it navigates with schemaId (should be with instanceId). Order of some logics and UI are not correct.


tuan-hda:
After successfully creating a new instance, the application incorrectly navigates to a view based on the schemaId associated with the instance type, rather than navigating to the specific, newly created instance using its unique instanceId. 

In the create instance function, logic of toasting is placed incorrectly. Text size of warning message when deleting schema needs to be smaller for consistency.

PR #340, opened 2026-03-06
Optimize loading time model page

trantienloi2404:
- Use promise.all to improve model listing api
- Add lazy loading PageModelList component
- Only display models for active tab
- Implement virtual scrolling in PageModelList

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,6 +38,7 @@
     "@rjsf/validator-ajv8": "^5.24.8",
     "@tanstack/react-query": "^5.37.1",
     "@tanstack/react-table": "^8.21.3",
+    "@tanstack/react-virtual": "^3.13.20",
     "@xyflow/react": "^12.4.3",
     "ajv": "^8.17.1",
     "axios": "^1.7.1",
--- a/frontend/src/components/molecules/DaSkeletonGrid.tsx
+++ b/frontend/src/components/molecules/DaSkeletonGrid.tsx
@@ -39,6 +39,7 @@ interface DaSkeletonGridProps {
   emptyContainerClassName?: string
   primarySkeletonClassName?: string
   secondarySkeletonClassName?: string
+  emptyAction?: React.ReactNode
   children?: React.ReactNode
 }
 
@@ -57,6 +58,7 @@ const DaSkeletonGrid = ({
   emptyContainerClassName = '',
   primarySkeletonClassName = '',
   secondarySkeletonClassName = '',
+  emptyAction,
   onTimeout,
   children,
 }: DaSkeletonGridProps) => {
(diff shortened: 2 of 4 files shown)


NhanLuongBGSV:
## PR #340 Review — Optimize loading time model page

**+652/-294 | 11 files | 4 commits**

### What it does

Performance optimization for the model listing page. Previously loaded all models from all three categories (owned, contributed, public) in a single request with up to 3000 items. Now:

**Backend:**
- `listAllModels` accepts `tab` param (`owned`/`contributed`/`public`) to fetch one category at a time
- Server-side pagination with `page` and `limit` params (default 24 per page)
- `Promise.all` for parallel stats computation (was sequential `for...of` loop)
- Backward compatibility: no `tab` param = legacy behavior (all three categories)
- Input validation added for `tab`, `page`, `limit`

**Frontend:**
- `useListModelsByTab` — infinite query hook, fetches one tab at a time with pagination
- `useModelTabCounts` — parallel queries to get total counts for tab badges
- `PageModelList` — complete rewrite: infinite scroll via `IntersectionObserver`, virtual scrolling via `@tanstack/react-virtual`, client-side search, lazy-loaded component
- `DaModelItem` wrapped in `React.memo()` with `loading="lazy"` on images
- `DaSkeletonGrid` — new `emptyAction` prop
- New dependency: `@tanstack/react-virtual@^3.13.20`

### Issues found

#### 1. Massive code duplication in backend (Major)

The `listAllModels` controller has the same stats-processing block copy-pasted 4 times (once for each tab case + legacy fallback). Each block is ~20 lines with identical `cacheResult` Map + `processStats` + `Promise.all` pattern. Should be extracted into a helper function called once per tab.

#### 2. Client-side search on partial data (UX gap)

Search only filters models already loaded in the client. If a user has 100 models and only 24 are loaded, searching for model #50 won't find it. The search input gives the impression of full-text search but only filters the current page. Should either:
- Add server-side search (a `q` or `name` query param)
- Or clearly indicate "searching loaded models only"

#### 3. Tab counts fire 3 separate API requests

`useModelTabCounts` makes 3 requests (`page=1, limit=1` each) just to get `totalResults` counts. Since `useListModelsByTab` already returns `totalResults` in the first page response, counts for the active tab could be derived from there — avoiding extra requests for inactive tabs until clicked.

#### 4. Virtual scrolling fixed row height

The virtualizer uses `estimateSize: () => 302` (fixed). If actual card heights vary (e.g., longer model names wrapping), the virtual list may have layout jumps. Consider using `measureElement` for dynamic sizing.

#### 5. `useListModelsByTab` enabled logic incomplete

```typescript
const enabled = !!apiTab && (activeTab !== 'myContribution' || !!self?.id)
```

Only gates `myContribution` on auth. The `myModel` tab (`created_by: req.user?.id`) will also fire for unauthenticated users — the backend returns empty gracefully but it's a wasted request. Should be `activeTab === 'public' || !!self?.id`.

#### 6. `retry` function usage in routes

```typescript
const PageModelList = lazy(() => retry(() => import('@/pages/PageModelList')))
```

`retry` is used but not shown in the diff. Verify it's imported in `routes.tsx`.

#### 7. ECA check failing

### Positive notes

- Sequential stats → `Promise.all` is a solid perf win
- `React.memo` on `DaModelItem` + lazy image loading are good optimizations
- Infinite scroll is much better UX than loading 1000 items upfront
- Backward-compatible API — no breaking changes for existing consumers
- Clean hook abstractions (`useListModelsByTab`, `useModelTabCounts`)

### Verdict

**Acceptable to merge with reservations.** The performance improvements are significant and well-structured. The backend code duplication (#1) should be cleaned up but isn't a blocker. The client-side-only search (#2) is a UX gap users may notice. The overall architecture is sound.

PR #343, opened 2026-03-06
Add Prototype configuration section

7inh:


NhanLuongBGSV:
## PR #343 Review — Add Prototype configuration section

**+376/-34 | 4 files | 1 commit**

### What it does

1. **New "Prototype Config" tab** in Site Config Management with dedicated `PrototypeConfigSection` component
2. **Migrated configs** from public to prototype category: `SHOW_CODE_API_PANEL`, `RUNTIME_SERVER_CONFIG` (moved), `SHOW_SDV_PROTOPILOT`, `SHOW_CODE_DIFF` (new)
3. **`PluginPageRender` refactoring:** `pluginAPI` memoized with `useMemo`, stable callback refs via `useRef`, `pluginConfig` memoized, dynamic `import()` before script injection fallback
4. **Re-introduced secret config exposure to plugins**

### Issues found

#### 1. CRITICAL: Secret configs re-exposed to plugins (Security regression)

This was fixed in PR #336 — `getPublicConfigs('site')` was used to only send public configs to plugins. This PR **reverts that fix**:

```typescript
Promise.all([
  configManagementService.getPublicConfigs('site'),
  configManagementService.getAllConfigs('site').catch(() => ({})),
]).then(([publicConfigs, allConfigs]) => {
  const secretConfigs: Record<string, any> = {}
  for (const [key, value] of Object.entries(allConfigs)) {
    if (!(key in publicConfigs)) {
      secretConfigs[key] = value
    }
  }
  setSiteConfigs({ public: publicConfigs, secret: secretConfigs })
})
```

This sends **all secret configs** (API keys, tokens, etc.) to every plugin via `config.secret`. Plugins are third-party code — they must never receive secrets.

#### 2. `SHOW_SDV_PROTOPILOT` key conflict with PR #338

PR #338 (already merged) added `SHOW_SDV_PROTOPILOT_BUTTON`. This PR adds `SHOW_SDV_PROTOPILOT` — a different key. The consumer in `PrototypeTabCode.tsx` (from PR #338) reads `SHOW_SDV_PROTOPILOT_BUTTON`. This new key won't be read by anything unless the consumer is also updated.

#### 3. Sequential deletes in factory reset

```typescript
for (const config of allConfigs.results || []) {
  try {
    if (config.id) {
      await configManagementService.deleteConfigById(config.id)
    }
  } catch (e) { ... }
}
```

Should use `Promise.allSettled()` or the shared `deleteConfigsById` utility from `siteConfigAdmin.ts` (added in PR #338).

#### 4. Code duplication with GenAIConfigSection

`PrototypeConfigSection` (241 lines) is another copy of the config section pattern. PR #338 extracted shared utilities in `siteConfigAdmin.ts` — this PR should use them instead of duplicating.

#### 5. `getAllConfigs` may require admin permissions

The code calls `configManagementService.getAllConfigs('site')` — if this maps to the admin-only config endpoint, non-admin users would get a 403 (silently caught, but secrets wouldn't load). Either way, secrets should not be fetched for plugins.

#### 6. ECA check failing

### Positive notes

- The `PluginPageRender` memoization improvements (useMemo for pluginAPI, useRef for stable callbacks, memoized pluginConfig) are solid performance wins — plugins won't re-render unnecessarily
- Dynamic `import()` with script injection fallback is a good pattern for plugin loading

### Verdict

**Do NOT merge.** The secret config re-exposure (#1) is a critical security regression that undoes the fix from PR #336. The config key mismatch (#2) means `SHOW_SDV_PROTOPILOT` won't control anything. These must be fixed before merging.

PR #352, opened 2026-03-19
Add Prototype Tab VSCode

trantienloi2404:
### Backend
- Add Coder service for workspace lifecycle (prepare, status, logs, timings)
- Add REST API: `/v2/system/coder/workspace/:prototypeId` (prepare, status, get, logs, timings)
- Add config: `CODER_URL`, `CODER_ADMIN_API_KEY`, `PROTOTYPES_PATH`

### Frontend
- Add `PrototypeTabVSCode` component embedding Coder workspace in iframe
- Add VS Code tab to prototype detail page
- Add `CoderWorkspaceStatus` for loading/error states

### DIFF
--- a/instance-setup/coder/workspace-image/home-seed/.local/share/code-server/User/settings.json
+++ b/instance-setup/coder/workspace-image/home-seed/.local/share/code-server/User/settings.json
@@ -54,5 +54,9 @@
     "**/node_modules": true,
     "**/bower_components": true,
     "**/*.code-search": true
+  },
+  "files.exclude": {
+    "**/.autowrx_out": true,
+    "**/.autowrx_run": true
   }
 }


NhanLuongBGSV (commented):
## Re-Review — New Commits (SHA: `1190f1d`)

This is a large PR (~3000+ lines) adding full Coder/VS Code workspace integration. Since there are new commits since my last review, here is an updated analysis.

### ✅ What's Good
- **Well-structured architecture**: Clean separation between `coder.service`, `orchestrator.service`, `gitea.service`, and `permissionSync.service`
- **Robust error handling**: Good use of `catchAsync`, proper HTTP status codes, and graceful fallbacks (e.g., session token generation tries multiple endpoints)
- **Race condition handling**: The duplicate workspace constraint recovery in `getOrCreateWorkspace` is thoughtful
- **Configuration via DB**: `coderConfig.js` reads settings from site-config with caching — avoids env-var sprawl

### 🔴 Critical Issues

1. **Weak password generation** (`coder.service.js:158`, `gitea.service.js:474`)
   `Math.random().toString(36)` is not cryptographically secure. Use `crypto.randomBytes()` instead:
   ```js
   const crypto = require('crypto');
   const generateRandomPassword = () => `pwd_${crypto.randomBytes(24).toString('hex')}`;
   ```

2. **Dead code: `gitea.service.js` + `permissionSync.service.js`**
   The Gitea integration is entirely disabled (every call in `orchestrator.service.js` is commented out), yet 780+ lines of dead code ship with this PR. This adds maintenance burden and attack surface. Either remove these files or gate them behind a feature flag in a separate PR.

3. **Missing auth on agent logs endpoint** (`coder.route.js:37`)
   `GET /workspaceagents/:workspaceAgentId/logs` only checks `auth()` but does NOT verify the requesting user owns or has access to that workspace agent. Any authenticated user can read any workspace agent's logs by guessing/enumerating UUIDs.

### 🟡 Medium Issues

4. **Massive code duplication in `coder.controller.js`**
   Every handler repeats the exact same pattern: fetch config → check enabled → find prototype → check permission. Extract a middleware like `requireCoderEnabled` + `requirePrototypeAccess(prototypeId)`.

5. **`getWorkspace` and `prepareWorkspace` are identical** (controller lines 19-50 vs 57-88)
   Both do the exact same thing. One should be removed or differentiated.

6. **Hardcoded UID/GID 1000** (`orchestrator.service.js:17-18`)
   `PROTOTYPES_LINUX_UID = 1000` / `PROTOTYPES_LINUX_GID = 1000` will break on systems where the container user has a different UID. Make this configurable via site-config.

7. **Overly permissive file modes** (`orchestrator.service.js:15-16`)
   `0o777` on directories and `0o666` on files is world-writable. Use `0o755`/`0o644` at minimum, or `0o770`/`0o660` if group access is needed.

8. **Agent-finding logic duplicated** (`coder.service.js`)
   The 5-method agent lookup pattern is copy-pasted in `getWorkspaceAppUrl` and `getWorkspaceAgentId`. Extract a shared `findWorkspaceAgent(workspace)` helper.

9. **Synchronous filesystem operations in async handler** (`orchestrator.service.js`)
   `fs.mkdirSync`, `fs.writeFileSync`, `fs.readdirSync` etc. in the request path block the event loop. Use async versions (`fs.promises`).

### 🔵 Minor Issues

10. **Copyright year**: Files say `2025` but we're in 2026. Update to current year.

11. **Logging sensitive data**: `getAdminHeaders()` token and `generateSessionToken` responses are logged at info level. Ensure tokens aren't leaked to log aggregators.

12. **`normalizeIdForName` defined twice**: Once in `coder.service.js` and once in `orchestrator.service.js`. DRY it out.

13. **Frontend `CoderWorkspaceStatus` component**: Not reviewed in detail but looks reasonable. Would benefit from error boundary wrapping for the iframe.

### Summary
The architecture is solid but there are security concerns (weak passwords, missing authz on logs endpoint, overly permissive file modes) and significant dead code that should be addressed before merge.

NhanLuongBGSV (commented):
## Re-review — New Commits (9091b0b)

This is a follow-up review after new commits were pushed. The PR has grown significantly with Coder/VS Code workspace integration. Here are my findings:

### 🔴 Critical

**1. World-writable file permissions (`0o777` / `0o666`)**
`orchestrator.service.js` sets `PROTOTYPES_DIR_MODE = 0o777` and `PROTOTYPES_FILE_MODE = 0o666`. This means any user/process on the host can read/write/execute prototype files. In a shared or multi-tenant environment this is a significant security risk. Consider using `0o755`/`0o644` at most, and relying on proper UID/GID ownership instead.

### 🟡 Medium

**2. Duplicate controller methods**
`getWorkspace` and `prepareWorkspace` in `coder.controller.js` are nearly identical — both call `orchestratorService.prepareWorkspaceForPrototype()` and return the same response shape. Consider consolidating into one endpoint or having `getWorkspace` be read-only (no creation side-effect).

**3. ~200 lines of commented-out Gitea code in `orchestrator.service.js`**
The entire Gitea flow (steps 2-4) is commented out with `//` blocks. This makes the file harder to read and maintain. If Gitea integration is disabled/deferred, remove the dead code and track it in a separate branch or issue.

**4. Sequential token endpoint fallback could be slow**
`generateSessionToken` in `coder.service.js` tries 3 different endpoints sequentially, catching 404s. On a slow Coder instance this could add significant latency. Consider probing once at startup and caching the working endpoint.

**5. Admin API key used for user-scoped token minting**
The admin key is used to create user-scoped tokens via `getAdminHeaders()`. While the tokens themselves are scoped, the admin key is called frequently. Consider caching the admin headers validation or adding rate limiting.

### 🟢 Minor

**6. Hardcoded UID/GID (1000:1000)**
`PROTOTYPES_LINUX_UID = 1000` and `PROTOTYPES_LINUX_GID = 1000` are hardcoded. This should be configurable via environment variables for deployments where the workspace container user has a different UID.

**7. `gitea.service.js` and `permissionSync.service.js` are added but unused**
These files (537 + 249 lines) are added to the codebase but all their consumers are commented out. This adds dead code to the bundle.

### 👍 What is good

- Proper permission checks in every controller method before workspace operations
- Clean separation: `coder.service.js` (Coder API) → `orchestrator.service.js` (business logic) → `coder.controller.js` (HTTP layer)
- Token scoping with allow-lists per workspace is a solid security pattern
- `$setOnInsert`-style logic (seed only if empty) for prototype files prevents data loss
- Good error handling with fallbacks (e.g., app URL not ready → return null, let FE poll)
- Comprehensive setup documentation in `instance-setup/coder/`

Overall a substantial feature addition. The main blockers are the `0o777` permissions and the dead code cleanup.

NhanLuongBGSV (commented):
## Review: PR #352 — Add Prototype Tab VSCode (Coder Integration)

This is a substantial PR (~3,600 lines across 30 files) introducing VS Code in-browser editing via Coder, plus a Gitea integration layer (currently disabled). The core idea is sound, but there are several critical issues that should be addressed before merging.

---

### 🔴 Critical

**1. `private: true` on Mongoose schema fields does nothing**

In `user.model.js`, sensitive fields like `coder_scoped_token`, `github_token`, `coder_scoped_token_expires_at`, and `coder_scoped_token_allow` are marked with `private: true`. This is not a standard Mongoose option — it is silently ignored and will NOT prevent these fields from appearing in API responses. If any query returns a User document without explicit field exclusion, tokens will leak.

Fix: Add explicit field exclusion in the schema's `toJSON` transform:
```js
transform: (doc, ret) => {
  delete ret.coder_scoped_token;
  delete ret.github_token;
  // etc.
}
```
Or add `select: false` to each sensitive field definition.

---

**2. `getWorkspace` endpoint allows write access with only READ permission**

In `coder.controller.js`, both `getWorkspace` and `prepareWorkspace` only check `PERMISSIONS.READ_MODEL`. However, a Coder workspace gives the user direct write access to prototype source files on the filesystem. This should require write (or contributor) permission, not just read.

---

**3. `model.reload()` does not exist in Mongoose**

In `permissionSync.service.js`, `syncPrototypeToGitea` calls `await model.reload()`. Mongoose documents have no `reload()` method — this will throw `TypeError: model.reload is not a function` at runtime. The intent seems to be to re-fetch the model after `syncModelToGitea` updates it; use `await Model.findById(model._id)` instead.

---

**4. Dead code shipped: Gitea is disabled but included**

`gitea.service.js` (537 lines) and `permissionSync.service.js` (249 lines) are fully committed but disabled via comments in `orchestrator.service.js`. Yet `permission.controller.js` still imports and calls `permissionSyncService.syncUserPermissionToGitea` and `removeUserPermissionFromGitea` — so Gitea sync is partially active (permission changes trigger it) while workspace creation does not. This creates an inconsistent state and the 786 lines of "disabled" services add significant maintenance burden. Either fully enable or fully remove before merging.

---

**5. Overly permissive file system modes**

`orchestrator.service.js` sets `PROTOTYPES_DIR_MODE = 0o777` and `PROTOTYPES_FILE_MODE = 0o666`. World-writable directories/files on a shared server allow any local user to read or overwrite prototype source code. Use `0o750` / `0o640` (or tighter) and set proper group ownership instead.

---

### 🟡 Medium

**6. `CODER_ADMIN_API_KEY` stored in site-config DB (admin UI)**

`coderConfig.js` reads the admin API key from site-config, which is accessible and editable by any admin through the UI. Admin API keys should be stored in environment variables / secrets management, not in a database that can be viewed/modified by multiple admins. This also means the key is exposed in any DB backup.

**7. `getWorkspace` and `prepareWorkspace` are redundant**

Both endpoints call `orchestratorService.prepareWorkspaceForPrototype()`. A GET endpoint (`getWorkspace`) should not have the side effect of creating a workspace. Consider splitting: GET returns current workspace state without creating, POST prepares/creates.

**8. `running` status shown as "Workspace is starting..."**

In `CoderWorkspaceStatus.tsx`:
```ts
running: 'Workspace is starting...',
```
`running` means the workspace IS running — the message should be `'Workspace is running'` or similar. This will confuse users who see a running workspace but the UI says it's "starting".

**9. No rate limiting on workspace prepare endpoint**

Repeated calls to `/workspace/:prototypeId/prepare` will spam the Coder API. Consider debouncing or adding a cooldown at the service level.

**10. `MAX_TOTAL_BYTES` scan budget in `prototype.service.js` is never enforced**

`listTextFilesRecursively` defines `MAX_TOTAL_BYTES = 5 * 1024 * 1024` but there is no accumulator checking total bytes across files. A prototype with thousands of small text files could cause unbounded memory usage.

**11. `coderConfig.getCoderConfigSync()` may return defaults on cold start**

On first request before the DB cache has been populated, `getCoderConfigSync()` returns `DEFAULTS` (enabled: false). This means early requests will get `403 VSCode integration is disabled` even if the admin has enabled it. The async `getCoderConfig({ forceRefresh: true })` is used correctly in controllers, but services that call `getCoderConfigSync()` directly could hit this window.

**12. No migration for new schema fields**

New fields added to `user.model.js`, `prototype.model.js`, `model.model.js` (`gitea_org_name`, `gitea_repo_url`, `coder_workspace_id`, etc.) have no migration notes. Existing documents will simply lack these fields (which is fine for optional fields), but it should be documented.

---

### 🟢 Minor

**13. Copyright year is 2025 in all new files** — the current year is 2026.

**14. `(timings as any)` cast in `CoderWorkspaceStatus.tsx`** — indicates the `WorkspaceTimings` type is incomplete. The `agent_script_timings` field should be added to the type definition in `coder.service.ts`.

**15. Gitea env vars in `.env.example` marked "DISABLED"** — confusing for new contributors. Either make them live or remove them.

**16. Missing tests** — None of the new services (coder, gitea, orchestrator, permissionSync) have accompanying test files.

---

### ✅ Positives

- Config-driven feature flag (`VSCODE_ENABLE` in site-config) with proper caching is a clean approach.
- Scoped token caching (`coder_scoped_token` + expiry) avoids minting a new token on every request — good for performance.
- `$setOnInsert` pattern reused from the instance snapshot PR — safe for idempotent workspace creation.
- Route structure, validation, and controller separation follow existing project conventions consistently.
- `PROTOTYPES_PATH` defaulting to a well-known path with configurable override is reasonable.
- Good use of `catchAsync` throughout controllers.

---

**Summary:** The VS Code integration concept is well-structured, but the dead Gitea code, the token leak risk via Mongoose schema, the incorrect permission check on workspace access, and the world-writable filesystem modes are blocking issues. Recommend addressing these before merge.

NhanLuongBGSV (commented):
## PR #352 — Add Prototype Tab VSCode (Coder Integration)

This is a substantial feature PR (~3,700 lines across 30 files) introducing Coder-based VSCode workspace support for prototypes. The architecture is thoughtful and the feature design is solid, but there are several issues that need addressing before merge.

---

### Critical

**1. Gitea code is "disabled" but still actively wired in**

`permissionSync.service.js` is imported and invoked in `permission.controller.js` for both `assignRoleToUser` and `removeRoleFromUser`. However, `orchestrator.service.js` comments out all Gitea/permissionSync imports as "DISABLED". This is inconsistent — Gitea sync code is simultaneously "disabled" and active. If Gitea is truly disabled, remove the calls from `permission.controller.js` too, or clearly document that it remains active. Currently this will make outbound HTTP calls to a Gitea instance that may not exist in deployments.

**2. `removeUserPermissionFromGitea` re-implements `getAuthHeaders` inline**

`permissionSync.service.js` `removeUserPermissionFromGitea()` re-defines `getAuthHeaders` and imports `axios`/`config` locally inside the function body — duplicating what `gitea.service.js` already exports. This will diverge silently over time. Use the exported helper instead.

**3. Scoped tokens stored as plaintext in MongoDB**

`coder_scoped_token` is persisted in the user document with `private: true`, but `private: true` is not a Mongoose encryption directive — it only signals intent (likely used by a custom `toJSON` transform to strip the field from API responses). The token is stored in cleartext. Given `github_token` has a comment `// Encrypted`, is there an encryption utility available that should be applied here too?

**4. `forceRefresh: true` on every controller request defeats the cache**

Every controller handler calls `getCoderConfig({ forceRefresh: true })` on every HTTP request, which triggers a fresh DB query every time. The caching mechanism (`CACHE_DURATION_MS = 10s`) in `coderConfig.js` is never used in practice. Either drop `forceRefresh: true` (rely on the 10s TTL) or document why per-request freshness is required.

---

### Medium

**5. Duplicate `sanitizePrototypeFolderName` function**

Defined identically in both `orchestrator.service.js` and `prototype.service.js`. Extract to a shared util.

**6. Duplicate `normalizeIdForName` function**

Defined in both `coder.service.js` and `orchestrator.service.js`. Same fix — shared util.

**7. `getWorkspace` and `prepareWorkspace` controllers are identical**

Both controllers perform the exact same operations (permission check + `prepareWorkspaceForPrototype`). If their purpose truly differs, make the distinction explicit in code. Otherwise consolidate.

**8. `PROTOTYPES_DIR_MODE = 0o777` is overly permissive**

Setting world-writable permissions (`0o777`) on workspace directories is a security concern in multi-tenant environments. Even if the container user needs write access, `0o755` or `0o770` with proper group ownership is safer. Hardcoded UIDs (`1000:1000`) also won't work in all deployment environments — consider making these configurable via site-config.

**9. `CoderWorkspaceStatus.tsx` casts timing objects to `any` throughout**

The component uses `(timings as any).provisioner_timings`, `(timings as any).agent_script_timings`, etc. in multiple places. The `WorkspaceTimings` type should include these fields properly so TypeScript can catch regressions.

**10. The `statusMessages` map shows "running" → "Workspace is starting..." which is misleading**

When the workspace status is `running`, the agent may not have completed startup yet. The message "Workspace is starting..." is shown for both `starting` and `running` states. The distinction between "workspace started but agent not ready" and "workspace is starting" is handled by `isCodeServerReady`, but the status message text itself is confusing.

---

### Minor

**11. `gitea.service.js` is 537 lines of dead code**

If Gitea is disabled, shipping a full service implementation creates maintenance burden and review surface for no benefit. Consider moving it to a feature branch or clearly marking the entire file with a top-level `// DISABLED` banner and documenting the intended re-enable path.

**12. `.env.example` includes disabled Gitea config without a clear deprecation note**

The example config adds 4 Gitea env vars labeled `(DISABLED - kept for reference)`. This will confuse operators. Either remove them or add a clear warning that they have no effect.

**13. No tests**

This is a significant backend feature with complex workspace lifecycle orchestration (prepare → status → logs → trigger-run → output). There are no unit or integration tests included. At minimum, the controller permission checks and `coderConfig` caching behavior should be tested.

**14. `coder_scoped_token_allow` type mismatch risk**

`getOrCreateUserScopedToken()` compares `String(user.coder_scoped_token_allow || '') === desiredAllow`. If `desiredAllow` is an empty string and `coder_scoped_token_allow` is undefined in the DB, this correctly returns `true`. But if it's `null` (Mongoose default for optional String), `String(null)` is `"null"` ≠ `""`. Add explicit null handling.

---

### Positives

- The `coderConfig` DB-backed config with caching is a clean approach — avoids requiring restarts for config changes.
- `$setOnInsert`-style seeding (avoid overwriting admin changes) mirrors the snapshot PR's design well.
- The scoped token mechanism (admin key mints user token, user token used for subsequent calls) is a good security layering approach.
- Permission checks before every workspace operation are consistently applied.
- File seeding from prototype code (including multi-file tree format) is well-handled.
- `CoderWorkspaceStatus` provides good progressive loading UX with live logs and timing breakdown.

PR #392, opened 2026-03-25
feat(#391): Instance Snapshot — export/import for reproducible deployments

NhanLuongBGSV:
Closes #391

## The Workflow

Deploy → Customize → Export → Redeploy

Fresh deploy uses code defaults. Admin configures everything via UI. When satisfied, clicks Export Snapshot — downloads a zip. Drop that zip into `./data/instance/` and the next deploy seeds from it automatically.

## What's in the Snapshot

```
instance-snapshot.zip
├── manifest.json
├── site-configs.json        (all key-values)
├── imgs/                    (logo, cover images)
└── seed/
    ├── plugins.json
    ├── model-templates.json
    └── dashboard-templates.json
```

## Key Design Decisions

Using `$setOnInsert` throughout — snapshot only seeds missing data, never overwrites what admin customized after deploy. Safe to redeploy without losing anything.

## Changes

- `instanceSnapshot.service.js` — export zip stream + startup seed logic
- `GET /v2/instance/export` — admin-only endpoint, streams zip
- `/instance/imgs/` served as static (instance logo overrides defaults)
- Startup: seeds predefined defaults → then seeds from instance bundle
- FE: Export Snapshot button in Site Management header
- docker-compose: `./data/instance:/usr/src/playground-be/instance` volume


NhanLuongBGSV (commented):
## PR Review — Instance Snapshot export/import

### Positives

- **`$setOnInsert` pattern** is the right choice — safe re-deploys that never overwrite admin customization
- **Secrets excluded** from export (`secret: false` filter on site configs)
- **Only user-uploaded images** are exported (checks for `/d/` path prefix), not baked-in defaults
- **Clean separation**: export as a streaming zip, seed logic reads from filesystem
- **Admin-only** route with proper `auth()` + `checkPermission(PERMISSIONS.ADMIN)`

### Issues & Suggestions

#### Security

1. **Filename injection** in `instance.route.js:17` — `req.query.name` is used directly in `Content-Disposition` header. A malicious admin could inject headers. Sanitize the `instanceName` (strip non-alphanumeric/dash characters) before using it in the header.

2. **Path traversal in seed** — `instanceSnapshot.service.js:213`: `imgValue.slice(3)` is joined directly with `STATIC_UPLOADS_DIR`. If a crafted `site-configs.json` in the bundle contains `../../../etc/passwd` as a value starting with `/d/`, `fs.existsSync` would check outside the uploads dir. Add a check that the resolved path stays within `STATIC_UPLOADS_DIR`.

#### Error Handling

3. **No error handler on archive** — If `archive.finalize()` or any DB query fails mid-stream, the response is already piped. Add `archive.on('error', ...)` to handle this gracefully.

4. **Startup seed errors don't prevent boot** — The `try/catch` blocks log errors but silently continue. This is probably intentional (don't block startup), but worth confirming. A partial seed could leave the instance in an inconsistent state.

#### Code Quality

5. **Repetitive seed logic** — The 4 seed blocks (configs, plugins, model templates, dashboard templates) follow an identical pattern. Consider extracting a helper like `seedCollection(filePath, Model, filterKey, extraFields)` to reduce duplication.

6. **Synchronous `fs.readFileSync`** in startup seed — Fine for startup, but could use `fs.promises` for consistency with the async pattern.

#### Frontend

7. **No user feedback on failure** — `handleExportSnapshot` catches errors with `console.error` only. Should show a toast/alert to the user.

8. **Emoji in button text** (`📦`, `⏳`) — Minor style nit, might not render consistently across all platforms. Consider using an icon component instead.

#### Docker

9. **Volume path** — `./data/instance` is added but won't exist by default. The seed logic handles the missing dir gracefully via the manifest check, but setup docs should mention it.

### Verdict

Solid feature with good design decisions. The **filename sanitization** (#1) and **path traversal** (#2) issues should be fixed before merge. The rest are improvements that could be addressed in follow-ups.

---

NhanLuongBGSV (commented):
## Code Review — Instance Snapshot (Export/Import)

**Overall:** Well-designed feature with clean idempotent seeding logic. The `$setOnInsert` approach is exactly right for reproducible deployments. A few issues to address before merge.

### 🟡 Medium

**1. Static route ordering contradicts comment (potential bug)**
In `app.js`, the comment says *"mount before /images so instance imgs can override defaults"* — but the route is mounted **after** `/images`:
```js
app.use("/images", express.static(...));
// Comment says "mount before /images" but appears AFTER
app.use("/instance/imgs", express.static(...));
```
Since these are different paths (`/images` vs `/instance/imgs`), Express won't conflict, so this is just a misleading comment. But if the intent is truly to *override* `/images` responses, the path should match or a middleware should intercept. Clarify the comment or the intent.

**2. No zip size limit on export**
`exportSnapshot` streams the entire `static/uploads/` directory into the zip. On a mature instance with lots of user uploads, this could produce a very large download. Consider:
- Adding a size cap or warning
- Only exporting admin-uploaded images (e.g., logo, cover) rather than **all** user uploads
- At minimum, document the expected size implications

**3. Synchronous file I/O in `seedFromInstanceBundle`**
The seed function uses `fs.readFileSync`, `fs.readdirSync`, `fs.copyFileSync` etc. For a startup-only path this is acceptable, but consider using async variants for robustness — especially the recursive `copyRecursive` which could block the event loop with many files.

### 🟢 Minor

**4. `assignAdmins` return value is fragile**
The change to return `(allAdmins[0]?.id || allAdmins[0]?._id) || null` couples the admin assignment logic to the snapshot seeding. If `assignAdmins` returns `null` (no admins configured), the `systemUserId` passed to seeders will be null, and the `...(systemUserId && {...})` spread silently skips `created_by`/`updated_by`. This is fine but worth a log line when it happens.

**5. No manifest version validation on import**
`seedFromInstanceBundle` reads the manifest but doesn't check `manifest.version`. Future snapshot format changes could break silently. A simple version check would be defensive.

### ✅ What's Good
- `$setOnInsert` / upsert pattern is exactly right — safe across restarts
- Clean separation between export (zip stream) and import (startup seed)
- Proper admin-only auth on the export endpoint
- Excellent documentation in the setup guide with the seed behavior summary table
- Frontend export UX is simple and effective

Solid PR overall — the medium issues are worth addressing but nothing is a blocker.

PR #420, opened 2026-03-31
Patch: Add asset and kit/runtime operations to PluginAPI

7inh:
This pull request adds comprehensive support for asset management and hardware kit/runtime operations to the `PluginPageRender` component and the `PluginAPI` interface. It introduces new methods for asset CRUD, sharing, and user search, as well as methods for interacting with hardware kits (fetching/replacing signal mapping and VSS files) via socket communication. These changes enable plugins to manage assets and interact with connected hardware kits in a standardized way.

**Asset Management Enhancements:**

- Added new asset operations to the `PluginAPI` interface, including methods to list, create, update, delete, share, and remove access to assets, as well as search users by email and get asset user lists. Each method is fully documented with usage examples.
- Implemented corresponding handler functions in `PluginPageRender.tsx` for each asset operation, with error handling and user feedback via toast notifications.
- Imported required services for asset operations and user search in `PluginPageRender.tsx`.
- Updated the `PluginAPI` object in `PluginPageRender.tsx` to expose the new asset operation handlers.
- Updated type imports to include `Asset` and `QueryAssetsParams` in `plugin.types.ts`.

**Hardware Kit / Runtime Operations:**

- Added new kit/runtime operations to the `PluginAPI` interface, including methods to fetch and replace signal mapping and VSS files on hardware kits, with detailed documentation.
- Implemented socket-based handler functions in `PluginPageRender.tsx` for fetching and replacing signal mapping and VSS files, including error handling and timeouts.
- Exposed the new kit/runtime operation handlers through the `PluginAPI` object in `PluginPageRender.tsx`.

### DIFF
--- a/frontend/src/components/organisms/PluginPageRender.tsx
+++ b/frontend/src/components/organisms/PluginPageRender.tsx
@@ -12,6 +12,10 @@ import { toast } from 'react-toastify'
 import { Spinner } from '@/components/atoms/spinner'
 import { getPluginById, getPluginBySlug } from '@/services/plugin.service'
 import { updateModelService, getComputedAPIs, getApiDetailService, replaceAPIsService } from '@/services/model.service'
+import config from '@/configs/config'
+import { io } from 'socket.io-client'
+import { listAssetsService, createAssetService, updateAssetService, deleteAssetService, shareMyAsset, removeUserFromShareList, getAssetById } from '@/services/asset.service'
+import { searchUserByEmailService } from '@/services/search.service'
 import { updatePrototypeService } from '@/services/prototype.service'
 import { listVSSVersionsService } from '@/services/api.service'
 import { uploadFileService } from '@/services/upload.service'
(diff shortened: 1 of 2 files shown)


NhanLuongBGSV (changes requested):
## Review: Request Changes

### Overall
Well-structured PR with good JSDoc documentation on types. Asset CRUD handlers are clean and consistent. However, the socket-based kit/runtime operations have several issues that should be addressed before merge.

---

### 🔴 Critical Issues

**1. Massive socket code duplication**

`handleFetchSignalMapping` and `handleFetchVss` are ~95% identical — only the file path differs. Same pattern for the replace operations. Extract a shared helper:

```ts
const socketReadFile = (kitName: string, filePath: string): Promise<string> => { ... }
const socketWriteFile = (kitName: string, filePath: string, content: string): Promise<void> => { ... }
```

Then each handler becomes a one-liner. This cuts ~120 lines of duplicated socket boilerplate.

**2. `handleReplaceVss` and `handleReplaceSignalMapping` have no timeout**

The fetch operations have a 10s timeout, but both replace operations can **hang forever** if the socket connects but never completes. Add a timeout (e.g. 15-20s) to all socket operations consistently.

**3. Hardcoded `setTimeout(3000)` for `generate_vehicle_model` is a race condition**

```ts
setTimeout(() => {
  socket.emit('messageToKit', { cmd: 'generate_vehicle_model', ... })
  finish(() => resolve())
}, 3000)
```

There is no guarantee the `write_file` command completes within 3 seconds. This should ideally wait for a `kitReply` confirming the write succeeded before triggering the rebuild. If that's not possible with the current kit protocol, at minimum document **why** 3s is used and make it a named constant.

---

### 🟡 Medium Issues

**4. No `disconnect` event handler on sockets**

If the socket drops unexpectedly (not `connect_error`), fetch promises hang until the 10s timeout, and replace promises hang forever (see #2). Add:
```ts
socket.on('disconnect', (reason) => {
  finish(() => reject(new Error(`Socket disconnected: ${reason}`)))
})
```

**5. No cleanup on component unmount**

Socket connections are created inside callbacks. If the component unmounts mid-operation (e.g. user navigates away), the socket stays alive and the promise/timeout leak. Consider using an `AbortController` or tracking active sockets with a ref for cleanup.

**6. Hardcoded identity: `username: 'plugin', user_id: 'plugin'`**

Every socket connection registers with the same static identity. Is this intentional? It means:
- No audit trail of which user/plugin triggered the operation
- Any plugin can interact with any kit

If this mirrors existing behavior elsewhere in the codebase, add a comment explaining why. Otherwise, consider passing the current user's identity.

---

### 🟢 Minor / Nits

**7. `payload: any` on socket responses** — Consider typing `kitReply` payloads for safety.

**8. Config paths as component-level constants** — `SIGNAL_CONFIG_PATH`, `VSS_PATH`, `KIT_SERVER_URL` are defined inside the component body (re-created every render). Move them outside the component or wrap in `useMemo`.

---

### ✅ What's Good
- Clean, consistent error handling pattern across all asset operations
- Excellent JSDoc with usage examples on every type definition
- Proper use of `useCallback` with correct dependency arrays
- Good PR description explaining the full scope of changes
- Asset operations are well-separated from kit operations

NhanLuongBGSV (approved):
## Review: Add asset and kit/runtime operations to PluginAPI

Clean, additive PR — 378 additions, 0 deletions. The asset operations wire up existing services correctly, JSDoc is thorough, and the type declarations in `PluginAPI` are well-structured.

**Approved with minor suggestions (non-blocking):**

### Duplicate socket scaffolding

`handleFetchSignalMapping`, `handleFetchVss`, `handleReplaceVss`, and `handleReplaceSignalMapping` all copy the same socket setup pattern (`settled`, `finish`, `timeout`). A small helper like `createKitSocket(kitName, kitServerUrl)` would reduce repetition significantly if this pattern grows. Not required for merge.

### Hardcoded credentials in `register_client`

All four handlers emit:
```js
socket.emit('register_client', { username: 'plugin', user_id: 'plugin', domain: 'domain' })
```
If the kit server uses these for anything beyond routing, they look like placeholder values. A brief comment explaining these are intentional (e.g. "kit server treats any 'plugin' client as anonymous read-write") would help future readers.

### Redundant `async` on `handleReplaceSignalMapping`

The outer function is declared `async` but immediately returns `new Promise()`. The `async` keyword is redundant here — the early `throw` works correctly because of the async wrapper, but it reads as if the function is doing something await-y at the top level that it isn't. Could just be a non-async function with `if (!model_id) return Promise.reject(...)`.

### `config.runtime?.signalConfigPath` / `vssPath` are not in the config type

These optional keys don't exist in `frontend/src/configs/config.ts`, so they always fall through to the hardcoded defaults. If they're meant to be configurable, they should be added to the config object. If not, the `config.runtime?.x ||` fallback is dead code.

---

None of these are blocking. The asset operations are correct and the socket kit API fills a genuine gap in plugin capabilities.

PR #431, opened 2026-04-02
Display default config key after restore default

7inh:


NhanLuongBGSV (approved):
## Review: Approve ✅

Small 7-line fix. Shows predefined config keys that are missing from the DB in the PublicConfigSection list, so admins can see all available default configs even before they are customized.

Clean approach using `filter` + `push` to merge DB configs with predefined defaults. LGTM.

PR #432, opened 2026-04-03
Refactor: dashboard template logic management and apply dashboard template logic

7inh:
This pull request introduces several improvements and refactors to both the backend and frontend dashboard template logic. The main changes include support for a new `default` dashboard template visibility mode, improved handling of template application logic in the dashboard UI, and enhancements to widget configuration based on selected signals. Additionally, some code cleanup and UI adjustments have been made.

**Dashboard Template Visibility and Validation:**

- Added a new `default` option to the `visibility` field in the `dashboardTemplate` schema, and updated validation logic to accept this value. This allows templates to be marked as default for auto-application. [[1]](diffhunk://#diff-c97261641e69b61bd537dec6ec54a55cba54080bbe3ce5a90dd29c3242d5e61eL16-R16) [[2]](diffhunk://#diff-f69fcaa874384b4cea1411b86149f77dd7cd10d8955c79410c253d8324d0119bL25-R25) [[3]](diffhunk://#diff-f69fcaa874384b4cea1411b86149f77dd7cd10d8955c79410c253d8324d0119bL37-R37)

**Template Application Logic and UI Behavior:**

- Refactored the logic in `DaDashboard.tsx` to auto-apply the template with `visibility: 'default'` if the user has not previously selected or customized a template. Improved the distinction between unset, null, and set template IDs.
- Removed legacy support for local templates and related UI, focusing template selection on backend-provided templates only. [[1]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L369-L385) [[2]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L92) [[3]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L175) [[4]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L184-L210)

**Widget Configuration and Signal Selection:**

- Updated widget processing logic so that, when signals are selected (`selected_signals`), they are automatically assigned to the `apis` field of widgets that do not already specify APIs. This is handled both when processing widgets and when editing/adding widgets in the editor. [[1]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L250-R250) [[2]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0R179-R182) [[3]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0R458-R462)

**Editor and UI Improvements:**

- Improved the dashboard editor by providing a fallback API list (parsed from `CVI_v4_1`) when no model is loaded, and cleaned up state management in the widget library/editor. [[1]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0R24-R37) [[2]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0R53-R69) [[3]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0L309-R333) [[4]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0L386-R416)
- Replaced a manual timeout with a more idiomatic `setTimeout`/`clearTimeout` for warning messages, and removed a redundant UI hint about clicking empty cells. [[1]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0L779-R809) [[2]](diffhunk://#diff-0380d53baebef24aca00e52ff5e5953844d00c693ce2147e99e0ad84175954f0L1106-L1110)

**Backend Service Update:**

- Optimized the dashboard template update service to validate only modified fields when saving, improving performance and reducing unnecessary validation.


NhanLuongBGSV (commented):
## Review: Comment (detailed)

Significant refactor of dashboard template logic — adds `default` visibility, inline editor, and fallback API support. 8 files, +142/-99.

### ✅ What's Good
- Adding `default` visibility tier makes template management more flexible
- Removing hardcoded `dashboard_templates` import in favor of DB-driven templates is the right direction
- `fallbackApis` prop on `ModelApiList` is a clean pattern for the editor use case
- Dashboard editor gets proper CVI parsing for API tree preview

### 🟡 Medium Issues

**1. `default` visibility semantics unclear**
The schema now allows `public | private | default` but there's no backend logic differentiating `default` from `public` in queries. Is `default` meant to be a system-provided template that can't be deleted? If so, add backend protection.

**2. `DashboardTemplateManager` grows significantly**
The inline widget config editor adds ~60 lines of JSX. Consider extracting `DashboardTemplateEditor` as a separate component if this page grows further.

**3. `DEFAULT_WIDGET_CONFIG` is `autorun: false`**
Is this intentional? Most dashboard templates probably want `autorun: true` to show widgets immediately.

### 🟢 Minor
- Clean removal of `dashboard_templates.ts` static data — good cleanup
- `activeModelApis` fallback logic with `?? []` is defensive, good

Overall solid refactor. No blocking issues.

NhanLuongBGSV (approved):
### 🔍 Code Review Summary

The PR implements a 'default' dashboard template system, allowing a specific template to be auto-applied when creating new prototypes. It also integrates the DaDashboardEditor into the template management flow and cleans up local/hardcoded template data.

#### ✅ Positive Changes
- **UX Improvement**: Moving the template editor to a dedicated component (DashboardTemplateEditor) that leverages the actual dashboard editor ensures consistency between template creation and prototype usage.
- **Robustness**: The addition of fallbackApis in ModelApiList using a parsed CVI allows the editor to function even when no specific vehicle model is loaded.
- **Code Cleanup**: Removal of hardcoded dashboardTemplates and the associated handleApplyLocalTemplate logic reduces tech debt.

#### ⚠️ Findings by Severity

**Minor**
- **Frontend Logic**: In DaDashboard.tsx, the check activeTemplateId === undefined is used to auto-apply the default template. Ensure that the API/database consistently returns undefined (or missing field) for new prototypes to avoid skipping this logic.
- **Styling**: The DashboardTemplateEditor uses w-[90vw] max-w-[1200px] h-[90vh]. Ensure this remains responsive on smaller screens or tablet views.
- **Formatting**: There are several small whitespace/formatting changes throughout DaDashboardEditor.tsx that are purely cosmetic.

**Overall**: The logic is sound and the implementation of the 'default' visibility state is correctly propagated from the Mongoose model through the Joi validations to the frontend types.

**Recommendation**: LGTM

mekong89:
## PR #432 Review: Refactor dashboard template logic

**Branch:** `fix/refactor-dashboard-template-logic` → `main`
**Author:** 7inh | **+296 / -218** across 11 files

---

### Summary
Adds a `default` visibility mode for dashboard templates that auto-applies when opening a prototype for the first time. Removes local/hardcoded templates in favor of backend-only templates. Extracts a new `DashboardTemplateEditor` component with inline widget layout editing. Also injects `selected_signals` into widget APIs automatically.

---

### Issues Found

**1. `visibility: 'default'` is semantically wrong**
"Default" is not a visibility concept — it's a behavioral flag (auto-apply). This conflates two concerns. A separate `is_default: Boolean` field would be cleaner and wouldn't break the existing public/private meaning. As-is, a "default" template is implicitly public, but that's not expressed anywhere.

**2. No uniqueness constraint on `default` visibility**
Nothing prevents multiple templates from being set to `visibility: 'default'`. `DaDashboard.tsx` uses `.find()` which silently picks the first one. The backend should enforce at most one default (e.g., unset other defaults on save).

**3. Bug: `processWidgetItems` destroys widget options**
In `DaDashboard.tsx`:
```js
if (!widget.options || !widget.options?.apis?.length) widget.options = {}
widget.options.apis = [...selectedSignals]
```
If a widget has options like `{ title: "Speed", apis: [] }`, this resets `widget.options` to `{}`, wiping `title` and all other properties. Should be:
```js
if (!widget.options) widget.options = {}
widget.options.apis = [...selectedSignals]
```

**4. Inconsistent signal injection logic across 3 locations**
- `processWidgetItems` — sets apis only when apis is empty/missing
- `DaDashboardWidgetEditor` — same (empty/missing check)
- `DaWidgetLibrary` — overwrites apis if it's **already an array** (`Array.isArray(options.apis)`)
- `DaWidgetList` — same as DaWidgetLibrary (overwrites existing)

These should behave consistently. The library/list variants will overwrite user-customized APIs, which is likely unintended.

**5. Inconsistent `autorun` default on parse failure**
In `DashboardTemplateEditor.tsx`:
- `DEFAULT_WIDGET_CONFIG` uses `autorun: false`
- The `catch` fallback in `save.mutationFn` uses `autorun: true`

**6. Dead code: `data/dashboard_templates` file not deleted**
The export was removed from `data/index.ts`, and the import was removed from `DaDashboard.tsx`, but the actual source file likely still exists on disk.

**7. Formatting-only changes add noise**
Multiple hunks in `DaDashboardEditor.tsx` are just re-indentation or whitespace changes. These should ideally be in a separate commit or excluded.

---

### Minor Notes
- `let` → `const` cleanup in `DaWidgetLibrary` is good
- `setTimeout`/`clearTimeout` replacing the async IIFE for warning messages is a nice fix
- The `hasTemplateDecision` logic using `in` operator to distinguish absent vs null is correct and well-commented
- Removing the "Click on empty cell" hint is fine

---

### Verdict
The core idea is sound, but there are a couple of bugs (especially #3 destroying widget options) and the `visibility: 'default'` design should be reconsidered before merging. Would recommend addressing at least issues #2, #3, and #4 before merge.

NhanLuongBGSV:
## PR #432 Re-review (updated diff)

**Previously reported issues — status:**
- ✅ `visibility: 'default'` — replaced with proper `is_default: Boolean` field
- ✅ No uniqueness constraint — backend now clears all other defaults before setting a new one in both `create` and `updateById`
- ✅ `processWidgetItems` destroys widget options — fixed, now only resets `widget.options` if it doesn't exist at all
- ✅ Inconsistent signal injection — all 4 locations now use the same `(!options.apis || options.apis.length === 0)` guard
- ✅ `autorun` fallback inconsistency — catch block now uses `false` to match `DEFAULT_WIDGET_CONFIG`
- ✅ Dead `dashboard_templates.ts` file — deleted

---

### Remaining Issues

**1. Race condition in uniqueness enforcement**
Both `create` and `updateById` do:
```js
await DashboardTemplate.updateMany({ is_default: true }, { is_default: false })
// ← gap here
return DashboardTemplate.create(body) // or doc.save()
```
Two concurrent requests with `is_default: true` can both clear the existing default, then both set themselves — ending up with two defaults. For an admin-only feature the risk is low, but worth noting. A proper fix would use a MongoDB transaction or a single atomic `findOneAndUpdate`.

**2. `handleSave` always fires now**
Old code: `if ((pendingChanges || hasTemplate) && prototype?.id)`
New code: `if (prototype?.id)`

Every Save click now saves unconditionally and always clears `dashboard_template_id`, even if nothing changed. Minor UX regression — clicking Save on an untouched prototype now overwrites the template link.

**3. Success toast removed on template apply**
```js
- toast.success(`Template "${template.name}" applied`)
```
Users now get no feedback when a template is applied. The error toast was kept but the success one was dropped. Should add it back.

---

### Verdict
Significantly improved — all structural issues from the first review are properly addressed. The 3 remaining items are minor: #1 is a low-risk edge case, #2 and #3 are small UX regressions. Good to merge with these noted.

PR #446, opened 2026-04-09
feat: restore global search feature

SangNT12:
## Summary
- Restore `DaGlobalSearch` component that allows users to search across Models and Prototypes from the navigation bar
- Search supports partial name/description matching (case-insensitive regex)
- Results are filtered based on user's access permissions

## Test plan
- [ ] Log in and verify Search button appears in the navigation bar
- [ ] Type a query and verify Models and Prototypes results appear
- [ ] Verify only accessible models/prototypes are returned


NhanLuongBGSV:
similar #448

PR #447, opened 2026-04-10
feat: add empty state component for vehicle model list

trantienloi2404:
- add empty state view for model list page


NhanLuongBGSV:
No need this PR, main is good now.

PR #462, opened 2026-04-13
Fix(#262): Should have option create empty prototype.

buiduyhieu1:
What have been changed:
- Add an option for empty prototype (Empty project template).
- Move all sample project templates to Database (adding along with predefined config).
- Create admin page for manage CRUD of project template.
- Allow user (with admin role) to save a Prototype as a project template.


NhanLuongBGSV (commented):
## Review

Good architecture overall — clean CRUD pattern consistent with `ModelTemplate`/`DashboardTemplate`, and the `$setOnInsert` seeding approach is correct (won't overwrite admin-modified templates). The Empty Project template properly addresses #262.

**Issue 1 — Duplicate `useEffect` in `FormCreatePrototype.tsx` (bug)**

The same `useEffect` block for initializing the first template is written twice in a row (lines ~200–215 in the diff). One copy should be deleted — as-is it runs the initialization logic twice on every `templateOptions` change.

**Issue 2 — `language` filter missing from list validation**

`projectTemplate.controller.js` picks `['name', 'language']` from the query, but the `list` schema in `projectTemplate.validation.js` only declares `name`. Any request filtering by `language` will silently have it stripped by the Joi validator before the controller sees it.

```js
// projectTemplate.validation.js — list query is missing language
query: Joi.object().keys({
  name: Joi.string(),
  // language: Joi.string(),  <-- missing
  sortBy: Joi.string(),
  limit: Joi.number().integer(),
  page: Joi.number().integer(),
}),
```

**Issue 3 — Raw `<input>` in Save Template dialog**

`PagePrototypeDetail.tsx` uses a raw `<input>` with inline Tailwind classes instead of the `<Input>` atom component — inconsistent with the rest of the codebase.

---

Please fix the duplicate `useEffect` (it's a real bug) and the validation gap before merge. The raw input is minor but worth cleaning up for consistency.

NhanLuongBGSV (changes requested):
## Review

Good overall structure — follows the ModelTemplate/DashboardTemplate pattern exactly, `$setOnInsert` seeding is correct (won't overwrite admin-modified templates), Joi validation is clean, and the admin-only write gate is correct. A few issues need fixing.

---

### 🔴 Backend: `language` filter field doesn't exist on the model

```js
// projectTemplate.controller.js
const filter = pick(req.query, ['name', 'language'])
```

The `ProjectTemplate` schema has no `language` field — language is stored *inside* the JSON `data` string. Passing `?language=python` to the list endpoint will silently return all templates unfiltered. Either remove the `language` filter, or restructure the model to have a top-level `language` field alongside `data`.

---

### 🟡 Frontend: missing `projectTemplate` in useEffect deps (FormCreatePrototype)

```tsx
useEffect(() => {
  if (templateOptions.length && !projectTemplate) {
    const first = templateOptions[0]
    setProjectTemplate(first.label)
    setData((prev) => ({ ...prev, code: first.code, language: first.language }))
  }
}, [templateOptions])  // ← projectTemplate is read here but missing from deps
```

React will warn about this (exhaustive-deps rule). Add `projectTemplate` to the deps array.

---

### 🟡 Frontend: no loading state while templates fetch

Both `FormCreatePrototype` and `FormNewPrototype` fetch templates via `useQuery` but show no loading indicator. In `FormCreatePrototype`, the Select shows a "Select a template" placeholder — users won't know if templates are still loading or genuinely empty. Add `isLoading` handling: disable the Select and show a spinner until templates are ready.

---

### 🟡 Frontend: handleSaveProjectTemplate doesn't invalidate queries

```tsx
const handleSaveProjectTemplate = async () => {
  // ...
  await createProjectTemplate(...)
  toast.success('Project template saved')
  setOpenSaveProjectTemplate(false)
  // ← queryClient.invalidateQueries missing
}
```

After saving, the `ProjectTemplateManager` list won't refresh until the next page load. Add:
```ts
queryClient.invalidateQueries({ queryKey: ['project-templates'] })
queryClient.invalidateQueries({ queryKey: ['project-templates-list'] })
```

---

### 🟡 Frontend: duplicate `Paged<T>` type — use existing `List<T>`

`projectTemplate.service.ts` defines a local `Paged<T>` interface that is identical to `List<T>` from `@/types/common.type.ts`. Replace with the shared type for consistency.

---

### Minor (non-blocking)

- `listProjectTemplates(params?: any)` — the `params` arg should be typed (e.g. `{ limit?: number; page?: number; name?: string }`) instead of `any`.
- `ProjectTemplateManager` uses raw `confirm()` for delete — the rest of the codebase uses `DaConfirmPopup`. Minor inconsistency.
- Both seeds in `index.js` are fire-and-forget (no `await`). Both have internal try/catch so this won't cause crashes, and it matches the existing pattern for `seedPredefinedSiteConfigs` — acceptable, but worth noting.

NhanLuongBGSV (approved):
All previously flagged issues have been resolved. Full re-review confirms this is ready.

**Fixed from the prior review:**
- `language` filter removed from `pick(req.query, ['name'])` — no phantom filter
- Duplicate `useEffect` eliminated — single init effect with correct deps `[templateOptions, projectTemplate]`
- Loading state added — spinner in Select + submit disabled during `isLoadingTemplates`
- Query invalidation complete — both `['project-templates']` and `['project-templates-list']` invalidated on every mutation
- `List<T>` from `@/types/common.type` used instead of local `Paged<T>`

**Everything else checks out:**
- `isAdmin` via `PERMISSIONS.MANAGE_USERS` is consistent with `NavigationBar`, `PluginManagement`, `PageManageFeatures`
- Route-level `checkPermission(PERMISSIONS.ADMIN)` (no resource ID) correctly falls through to global role check in `hasPermission`
- `$setOnInsert` seeding correct — admin-modified templates are never overwritten on restart
- Predefined templates use `JSON.stringify()` matching the `data: String` schema field
- `sampleProjects.ts` fully deleted and removed from `data/index.ts`
- `/admin/project-templates` route placement consistent with `TemplateManager` / `DashboardTemplateManager`

Approved to merge.

PR #465, opened 2026-04-14
Fix auth flicker on home

### LINKED ISSUE #464: Login state flickers on home and emits initial 401 requests after successful sign-in
## Problem
After a user signs in successfully and navigates to the home page, the UI briefly shows the logged-out state (for example, `Sign In`) before switching back to logged-in state.
At the same time, the browser console/network shows initial `401` responses from:
- `GET /users/self`
- `GET /permissions/has-permission`
## Impact
- Visible UX glitch (“auth flicker”) that reduces trust in login state.
- Noisy console/network logs with transient unauthorized requests.

trantienloi2404:
Fixes #464

- Gate self-profile and permission queries until auth is bootstrapped and access token exists.
- Prevent UI from briefly showing logged-out state while auth is still resolving.


NhanLuongBGSV (commented):
## Review

The bootstrap approach is solid — gating queries behind `authBootstrapped` cleanly fixes the flicker, and replacing the page-reload on login/register with direct state updates is a good improvement. Two things need to be addressed before merge:

---

### Issue 1 — 401 auto-refresh removed with no replacement (major)

`QueryProvider` drops the `QueryCache.onError` handler that refreshed the token and retried queries on 401:

```ts
// Removed from QueryProvider:
queryCache: new QueryCache({
  onError: async (error, query) => {
    if (isAxiosError(error) && error?.response?.status === 401) {
      await refreshAuthToken()
      queryClient.invalidateQueries({ queryKey: query.queryKey })
    }
  },
}),
```

Access tokens are typically short-lived (15 minutes). The bootstrap only fires once on mount. Once the token expires mid-session, every API call returns 401 with no recovery mechanism — the user would need to manually reload the page to get a new token.

**What to do:** Keep the `QueryCache.onError` 401 handler alongside the new bootstrap. The two approaches are complementary — bootstrap handles the initial load, the 401 handler handles mid-session expiry. You can simplify `refreshAuthToken` to just call `/auth/refresh-tokens` (same endpoint used in bootstrap) and call `setAccess` on success:

```ts
queryCache: new QueryCache({
  onError: async (error, query) => {
    if (isAxiosError(error) && error?.response?.status === 401) {
      try {
        const res = await refreshAxios.post('/auth/refresh-tokens', {})
        if (res.data?.access?.token) {
          setAccess(res.data.access)
          queryClient.invalidateQueries({ queryKey: query.queryKey })
        }
      } catch {
        logOut()
      }
    }
  },
}),
```

---

### Issue 2 — Merge conflict on `NavigationBar.tsx`

PR #463 (refactor tools menu) was merged today and rewrote `NavigationBar.tsx`. This PR's diff is based on the old version so it will conflict.

**What to do:** Rebase this branch on `main` and resolve the conflict. The changes are compatible — you just need to apply the `isResolvingAuth` logic on top of the refactored file.

NhanLuongBGSV (changes requested):
Great progress — the rebase on main is clean and restoring the 401 handler in QueryProvider is exactly right. One remaining issue before merge:

### Missing concurrent refresh guard (regression from old code)

The old `QueryProvider` had a `refreshingToken` ref to prevent multiple simultaneous refresh calls:

```ts
if (refreshingToken.current) return
try {
  refreshingToken.current = true
  // ... refresh
} finally {
  refreshingToken.current = false
}
```

The new `QueryCache.onError` handler doesn't have this guard. If 3 queries all return 401 at the same time, all 3 fire `POST /auth/refresh-tokens` concurrently. The first succeeds; the second and third hit a consumed refresh token, land in `catch`, and call `logOut()` — logging the user out even though the token was just successfully refreshed.

Fix: add a module-level (or component-level via `useRef`) flag to deduplicate:

```ts
let isRefreshing = false

// inside onError:
if (isAxiosError(error) && error?.response?.status === 401) {
  if (isRefreshing) return
  isRefreshing = true
  try {
    const res = await refreshAxios.post('/auth/refresh-tokens', {})
    if (res.data?.access?.token) {
      setAccess(res.data.access)
      query.invalidate()
    }
  } catch {
    logOut()
  } finally {
    isRefreshing = false
  }
}
```

Everything else looks good — the bootstrap flow, the `authBootstrapped` guard on `useSelfProfile` / `usePermissionHook`, and the NavigationBar changes are all correct.

NhanLuongBGSV (commented):
## Deep Review: Fix auth flicker on home

This is a thorough, well-structured solution. The bootstrap pattern is correct and the `authBootstrapped` flag is wired consistently. Here's the full breakdown:

---

### Architecture — correct

The three-phase auth flow is clean:
1. **Bootstrap** (`RootLayout` effect): fire `POST /auth/refresh-tokens` once, set `authBootstrapped` in `.finally()` regardless of outcome
2. **Query gate** (`useSelfProfile` + `usePermissionHook`): only run once `authBootstrapped && !!accessToken`
3. **UI gate** (`isResolvingAuth`): show placeholders until auth is known

The `bootstrappingRef` double-invocation guard, the `DaNavUser` fixed-size placeholder (`h-9 w-20`), and the `useEffect(..., [user?.id])` dep fix in `HomePrototypePopular` are all good.

---

### Issue 1 — BLOCKING: concurrent refresh deduplication (already in review above)

Already flagged. The old `refreshingToken.current` guard must be restored.

---

### Issue 2 — `authResponse: any` loses type safety in FormSignIn/FormRegister

```ts
const authResponse: any = await loginService(email, password)
if (authResponse?.user && authResponse?.tokens?.access) {
```

`loginService` is typed as returning `AuthToken` which has `access: Token` at the top level, but the actual API response has the shape `{ user, tokens: { access, refresh } }`. The `any` cast works but bypasses TypeScript entirely. The real fix is to define a proper type:
```ts
type LoginResponse = {
  user: User
  tokens: { access: Token; refresh: Token }
}
```
and update `loginService` / `registerService` return types. Low priority but cleans up the confusing `AuthToken` type that has both `access: Token` and `tokens: any`.

---

### Issue 3 — `refreshAxios` created on every render of QueryProvider

```ts
const QueryProvider = ({ children }) => {
  const refreshAxios = axios.create({...})  // ← recreated every render
  const [queryClient] = useState(() => new QueryClient({
    queryCache: new QueryCache({
      onError: async (error, query) => {
        // closes over refreshAxios from *first* render
```

The `queryClient` is created once; `onError` captures the `refreshAxios` from the first render. Subsequent `refreshAxios` instances are created but never used. Harmless in practice (first-render instance is valid), but should be moved inside the `useState` initializer or lifted to module scope.

---

### Issue 4 — No page reload after login: stale state risk

The switch from `window.location.href = window.location.href` to in-memory state update is a welcome improvement — no flash, no lost router state. However, any component that previously relied on the full page reload to reset its own local state may now retain stale data after login. Worth a smoke-test pass on pages with protected content.

---

### Note — `logOut()` still navigates away

`logOut()` calls `logoutService().then(() => window.location.pathname = '/')`. If the concurrent-401 race condition fires before Issue 1 is fixed, multiple `logOut()` calls means multiple page navigations queued. Not harmful but worth keeping in mind.

---

### Note — `authBootstrapped` dep in bootstrap useEffect

```ts
useEffect(() => { ... }, [authBootstrapped, setAccess, setAuthBootstrapped])
```

When bootstrap completes, `authBootstrapped` flips to `true`, the effect re-runs, and immediately returns. Correct behaviour, just two runs instead of one. Empty deps `[]` with an internal `useRef` check would be equivalent and cleaner — but the current approach is not wrong.

---

### Summary

Fix Issue 1 (concurrent guard) and Issue 2 (type) before merge. Issues 3 and 4 are follow-up polish.

NhanLuongBGSV (approved):
All blocking issues resolved in the latest commit:

- **Concurrent refresh guard**: `isRefreshing` and `refreshAxios` are now module-level — concurrent 401s are correctly deduplicated ✅
- **Type safety**: `authResponse: any` removed, `loginService`/`registerService` explicitly typed as `Promise<AuthToken>` ✅  
- **`refreshAxios` on every render**: eliminated by lifting to module scope ✅
- **Bonus**: `registrationData: any` replaced with a proper typed object, `AuthToken` updated with `user: User` and `tokens: { access, refresh }` ✅

Minor note: `AuthToken` now covers both the login shape (`user + tokens`) and the refresh shape (`access` top-level) in one type — technically two shapes in one, but works correctly with optional chaining everywhere it's used. Worth splitting into `LoginResponse` / `RefreshResponse` as a follow-up if desired.

Ready to merge.

PR #471, opened 2026-04-15
Add disscuss feature into api detail

7inh:
This pull request implements a complete discussion system in the frontend, replacing the previous stub. It introduces new components, hooks, and services to support listing, creating, updating, and deleting discussions and replies. The discussion UI is now fully functional, including nested replies, editing, deleting with confirmation, and user avatars. Supporting changes include updates to the data types, service layer, and integration into the API detail and home pages.

**Discussion System Implementation**

- Implemented a fully functional discussion UI in `DaDiscussions.tsx`, supporting listing, nested replies, editing, deleting (with confirmation), and user avatars. The component uses new hooks and services for data fetching and actions.
- Added the `useListDiscussions` hook to fetch paginated discussions for a given reference and type, leveraging React Query for caching and reactivity.
- Replaced the stubbed discussion service with real API calls for listing, creating, updating, and deleting discussions in `discussion.service.ts`.

**Type and Data Structure Updates**

- Updated the `Discussion` type to include a `created_by` object (with id, name, and optional image) instead of just a user id, and added an optional `replies` array for nested discussions.
- Modified the mock data in `discussion_mock.ts` to match the new `created_by` structure. [[1]](diffhunk://#diff-0c755fcab41ae2a2d46b138f9c453f907c6caa843458d6fcef9347962599869cL17-R17) [[2]](diffhunk://#diff-0c755fcab41ae2a2d46b138f9c453f907c6caa843458d6fcef9347962599869cL27-R27) [[3]](diffhunk://#diff-0c755fcab41ae2a2d46b138f9c453f907c6caa843458d6fcef9347962599869cL37-R37)

**UI and Integration Enhancements**

- Integrated the new discussion system into the API detail page, ensuring proper layout and width handling.
- Improved the discussion creation form to prevent empty submissions and fixed update logic to only update the content field. [[1]](diffhunk://#diff-acc0b26fe6b4e09b05fd5995af9d32f678030544207fa8ec2e52d2b2eb30c88aL130-R130) [[2]](diffhunk://#diff-acc0b26fe6b4e09b05fd5995af9d32f678030544207fa8ec2e52d2b2eb30c88aL164-R164)
- Added a dialog on the home page that can be triggered via a URL parameter for creating a model, improving navigation and user experience. [[1]](diffhunk://#diff-a497e1baec5dd2d81bab57632bc5f06f247a2f8a6aca5369e767c085662f50a2R10) [[2]](diffhunk://#diff-a497e1baec5dd2d81bab57632bc5f06f247a2f8a6aca5369e767c085662f50a2R21-R35) [[3]](diffhunk://#diff-a497e1baec5dd2d81bab57632bc5f06f247a2f8a6aca5369e767c085662f50a2R96-R109)


NhanLuongBGSV (approved):
## Review: Add discussion feature into API detail

Solid implementation — replaces the stub with a real discussion UI including nested replies, edit/delete with confirmation, empty state, and proper service layer. The type changes (DiscussionUser, replies?: Discussion[]) are correct. FormCreateDiscussion fix (only copying content on edit, disabling submit on empty) is good.

**Minor suggestions (non-blocking):**

### Unrelated change in PageHome.tsx

The ?create-model URL-param feature has nothing to do with discussions. Bundling it here makes it harder to track via git history. Ideally it lives in its own PR, but won't block merge.

### Delete confirmation truncation

If discussion.content is long this produces an unwieldy confirmation string. Truncate to ~80 chars before injecting into the label.

### "Reply (n)" button label is misleading

The button labeled Reply (count) actually *expands* the reply list, not creates a reply. Renaming it to "Replies (n)" would remove the ambiguity.

### No error state

useListDiscussions can return an error but the component only handles isLoading and the empty case. A simple error branch would prevent a silent blank screen on API failure.

### No pagination UI

useListDiscussions accepts page but the component always passes page 1. Fine for now but worth noting as the feature grows.

PR #472, opened 2026-04-15
Enable prototype renaming via button

7inh:
This pull request updates the `DaPrototypeItem` component to improve usability and clarify access to prototype actions. The changes ensure that only logged-in users see interactive features like renaming, and context menu options are gated behind a feature flag. The most important changes are:

**User access and UI controls:**

* The prototype name is now rendered as a button with an edit (pencil) icon for logged-in users, allowing them to open the rename dialog directly by clicking the name. For non-logged-in users, the name remains plain text.
* The context menu and its associated dialogs (deploy/staging and delete) are only available to logged-in users, and are further gated by the `enableContextMenu` flag. [[1]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dL274-R293) [[2]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dR305) [[3]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dR417-R418) [[4]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dR441-R442)

**Dialog and interaction improvements:**

* The rename dialog can now be triggered both from the context menu and the new pencil icon button, making it more accessible.

These changes clarify which actions are available to which users and streamline prototype management interactions.

### DIFF
--- a/frontend/src/components/molecules/DaPrototypeItem.tsx
+++ b/frontend/src/components/molecules/DaPrototypeItem.tsx
@@ -243,9 +243,28 @@ const DaPrototypeItem = ({ prototype, className }: DaPrototypeItemProps) => {
           </div>
         </div>
         <div className="flex items-center w-full space-y-0">
-          <p className="text-base font-semibold line-clamp-1 text-foreground prototype-grid-item-name">
-            {prototype?.name ?? ''}
-          </p>
+          {user ? (
+            <button
+              type="button"
+              className="flex items-center gap-1 min-w-0 text-left cursor-pointer group/rename"
+              onClick={(e) => {
+                e.stopPropagation()
+                e.preventDefault()
+                setNewName(prototype?.name ?? '')
+                setRenameOpen(true)
+              }}
+              aria-label="Rename prototype"
+            >
+              <span className="text-base font-semibold line-clamp-1 text-foreground prototype-grid-item-name">
+                {prototype?.name ?? ''}
+              </span>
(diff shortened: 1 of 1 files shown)


NhanLuongBGSV (approved):
## Review: Enable prototype renaming via button

Clean refactor of a single file. The pencil-icon-on-hover pattern is a good UX improvement, touch support via pointer-coarse is thoughtful, and correctly splitting the gate so Deploy/Delete stay behind enableContextMenu while Rename becomes always-available for logged-in users makes sense.

**Two existing issues the PR exposes more prominently:**

### Rename button shown to non-owners

The pencil icon is gated on any logged-in user (user != null), not on ownership. A user who doesn't own the prototype will see the rename button, open the dialog, submit, and hit a 403 from the backend. The dialog stays open with no feedback (see below). To fix, gate the button on ownership:

```tsx
const isOwner = user?.id === prototype?.created_by?.id
// ...
{isOwner ? <button ...rename button...> : <p ...plain name...>}
```

### Silent failure on rename error

`handleRename` catches errors but only calls `console.error`. No toast or inline message is shown to the user — the dialog just stays open. A non-owner hitting 403 (or any network error) gets zero feedback. Add a toast in the catch block.

---

Neither of these was introduced by this PR (the rename dialog existed before), but the new pencil icon makes non-owner failure more likely. Approving since the core change is correct — suggest fixing both in a follow-up.

PR #473, opened 2026-04-15
Add gradient header support to DaDashboard component

7inh:
This pull request updates the `DaDashboard` component to add support for an optional gradient header in fullscreen mode, based on a new site configuration setting. The most important changes are:

**UI Theming Enhancements:**

* Added support for a `GRADIENT_HEADER` site config option, enabling a gradient background and adjusted text color for the dashboard header in fullscreen mode. [[1]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7R56) [[2]](diffhunk://#diff-ba057e561595cb29932fc1af30b1335b6bf78f8936d3d432561df529316ab0e7L315-R338)
* Updated the logo (`DaImage`) to invert its colors when the gradient header is active, ensuring visibility against the new background.

**Configuration Management:**

* Utilized the `useSiteConfig` hook to retrieve the new `GRADIENT_HEADER` setting, defaulting to `false` if not set.

### DIFF
--- a/frontend/src/components/molecules/dashboard/DaDashboard.tsx
+++ b/frontend/src/components/molecules/dashboard/DaDashboard.tsx
@@ -53,6 +53,7 @@ import { useSiteConfig } from '@/utils/siteConfig'
 const DaDashboard = () => {
   const { data: model } = useCurrentModel()
   const logoUrl = useSiteConfig('SITE_LOGO_WIDE', '/imgs/logo-wide.png')
+  const gradientHeader = useSiteConfig('GRADIENT_HEADER', false)
   const [
     prototype,
     setActivePrototype,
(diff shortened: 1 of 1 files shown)


NhanLuongBGSV (approved):
Clean, minimal feature flag addition. Reads the existing GRADIENT_HEADER config (default false, no new config required) and applies the gradient only in fullscreen mode — no change to default behavior.

Two tiny observations (non-blocking):

1. `showPrototypeDashboardFullScreen && gradientHeader ? '' : 'bg-white'` passes an empty string to cn() which is harmless, but `!(showPrototypeDashboardFullScreen && gradientHeader) && 'bg-white'` reads more clearly.

2. `brightness(0) invert(1)` forces the logo fully white, which works for dark logos but could look off if a site configures a light/colorful logo intended for dark backgrounds. Worth noting in case logo support needs to be more flexible in future.

Otherwise correct — inline style for CSS variables is the right approach, and the gradient only activates when both the fullscreen state and the config flag are true.

PR #477, opened 2026-04-16
Enhance prototype owner control with context menu permissions

7inh:
This pull request updates the `DaPrototypeItem` component to enforce ownership-based permissions for editing and context menu actions. Only the owner of a prototype or the associated model can access editing features and the context menu. Non-owners attempting to use the context menu will now receive a notification instead of being able to interact with edit options.

**Permission Enforcement and User Feedback:**

* Added an `isOwner` check to determine if the current user is the owner of the prototype or its parent model, and updated all relevant UI logic to use this check.
* Restricted edit buttons and context menu access to owners only; non-owners will see a notification via `react-toastify` if they attempt to use the context menu. [[1]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dL246-R252) [[2]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dL305-R311) [[3]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dR375-R385) [[4]](diffhunk://#diff-8b9feb56fe8809f88cf179e9057883fed171f585014f05087f9bd786f14c968dL417-R434)
* Updated the UI to hide edit-related controls from non-owners, ensuring a clearer and more secure user experience.

**Dependency Update:**

* Imported the `toast` function from `react-toastify` to provide user notifications for permission errors.

### DIFF
--- a/frontend/src/components/molecules/DaPrototypeItem.tsx
+++ b/frontend/src/components/molecules/DaPrototypeItem.tsx
@@ -51,7 +51,7 @@ import { Input } from '../atoms/input'
 import useCurrentModel from '@/hooks/useCurrentModel'
 import useListModelPrototypes from '@/hooks/useListModelPrototypes'
 import PrototypeTabStaging from '@/components/organisms/PrototypeTabStaging'
-import { toast } from 'react-toastify'
+import { useToast } from '@/components/molecules/toaster/use-toast'
 
 interface DaPrototypeItemProps {
   prototype?: Prototype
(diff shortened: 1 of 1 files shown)


NhanLuongBGSV (approved):
Clean, targeted fix that directly addresses the ownership gap identified in the PR #472 review.

**What this does right:**
- `isOwner` logic is correct: `!!user && (user.id === prototype?.created_by?.id || user.id === model?.created_by?.id)` — covers both prototype-level and model-level ownership, consistent with the rest of the codebase
- Rename button (pencil icon) now properly gated on `isOwner` instead of just `user` — eliminates the silent 403 that non-owners would have hit
- Context menu gated on `enableContextMenu && isOwner` — no more context menu leaking to non-owners
- Toast notification for right-click permission denial is a good UX choice (clear feedback instead of silent no-op)
- View Code/Dashboard quick-links condition `!isOwner` is semantically correct (show navigation shortcuts when the user can't edit)
- Deploy and Delete dialogs also gated on `enableContextMenu && isOwner` — consistent

**No concerns.** Approved to merge.

PR #499, opened 2026-06-01
Change Architecture section header to H3

yamini178:


NhanLuongBGSV:
Hi @yamini178, thanks for the contribution! A few questions before we can merge this:

1. **PR title vs actual changes**: The title says "Change Architecture section header to H3" but the PR also includes a full rewrite of `HomeHeroSection.tsx` (gradient overlay system, configurable image/text positioning) and changes to `PageModelList.tsx`. Could you update the PR title and description to reflect all the changes?

2. **Auth gate on model list**: `PageModelList.tsx` now shows a "Sign in required" lock screen to unauthenticated users instead of the public model list. Was this intentional? Previously unauthenticated users could browse public models — this is a behavioral change that could break public access. Can you explain the intent here?

jayantadn:
> Hi @yamini178, thanks for the contribution! A few questions before we can merge this:
> 
> 1. **PR title vs actual changes**: The title says "Change Architecture section header to H3" but the PR also includes a full rewrite of `HomeHeroSection.tsx` (gradient overlay system, configurable image/text positioning) and changes to `PageModelList.tsx`. Could you update the PR title and description to reflect all the changes?
> 2. **Auth gate on model list**: `PageModelList.tsx` now shows a "Sign in required" lock screen to unauthenticated users instead of the public model list. Was this intentional? Previously unauthenticated users could browse public models — this is a behavioral change that could break public access. Can you explain the intent here?

@yamini178 : requesting you to change the PR title and description to meet the actual changes. Please follow this as a standard practice to give proper headline and description for any PR you raise.

@NhanLuongBGSV : The background of the change is that, when user is trying to browse ORCA without login, an unknown error was thrown. I agree for digital auto, browsing publilc models without sign in is allowed. Do we have the same strategy for ORCA?

NhanLuongBGSV:
@jayantadn This setting already exists and can be changed on the Site Settings page, avoiding the need to reimplement it. ORCA contains private Bosch data, so users must log in to view it.

jayantadn:
@NhanLuongBGSV : Exactly, so model viewing without login should not be allowed. Hence the change introduced by Yamini. I guess now we have to find a way to distinguish between orca and digital.auto. @yamini178 : can you please explore this a bit?

NhanLuongBGSV:
That is a misunderstanding. You are trying to implement a feature that already exists.

jayantadn:
ok, then let us discuss next week, when Yamini is in office.

PR #501, opened 2026-06-08
Feat: Remove manage hardware kit hard code kit.digitalauto.tech

7inh:
This pull request introduces a new "forceKitId" feature to the `DaRuntimeConnector` component, allowing the UI to track and display the status of a specific runtime (kit), even if it's offline or doesn't match the usual filtering. It also improves configuration flexibility by using the `useSiteConfig` hook to set the runtime server URL in several places. The main changes are grouped below:

### ForceKitId Feature and Runtime Status Display

* Added a new optional `forceKitId` prop to `DaRuntimeConnector`, enabling the component to always include and show the status of a specific kit, regardless of its online status or prefix. The UI now displays a status indicator ("Connected", "Disconnected", "Unreachable", or "Connecting...") for the forced kit. [[1]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR34) [[2]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR54) [[3]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR393-R410) [[4]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR621-R656)
* Updated the runtime filtering logic so that when `forceKitId` is set, all runtimes are included (not just online ones), and prefix filtering is bypassed.
* Added and managed a `hasLoadedKitList` state to help determine when the kit list is loaded for accurate status display. [[1]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR77) [[2]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR364) [[3]](diffhunk://#diff-e0d8b62484e9acf8ee1e4e5049e9e11b1815b088029f543a8ac93f8d637d6c2aR393-R410)

### Configuration Improvements

* Replaced direct usage of the `config.runtime.url` value with the `useSiteConfig` hook in both `FormHardwareKitManager` and `PluginPageRender`, making the runtime server URL configurable via site settings. [[1]](diffhunk://#diff-8d414cc9c18100c04f4a103c106c05fd162b34cebabdcca81c238e60ff7481c8R10) [[2]](diffhunk://#diff-8d414cc9c18100c04f4a103c106c05fd162b34cebabdcca81c238e60ff7481c8R205) [[3]](diffhunk://#diff-8d414cc9c18100c04f4a103c106c05fd162b34cebabdcca81c238e60ff7481c8L245-R249) [[4]](diffhunk://#diff-da6ff398bcba7362914a94ffae896432694b4bb62ec5b21c50f07483195cafc6R37) [[5]](diffhunk://#diff-da6ff398bcba7362914a94ffae896432694b4bb62ec5b21c50f07483195cafc6R69) [[6]](diffhunk://#diff-da6ff398bcba7362914a94ffae896432694b4bb62ec5b21c50f07483195cafc6L356-R358)

### Miscellaneous

* Added a debug log for the kit server URL in `DaRuntimeConnector` for easier troubleshooting.

These changes collectively make the runtime connection UI more robust and configurable, especially for scenarios where a specific kit must be tracked regardless of its online status.

### DIFF
--- a/frontend/src/components/molecules/DaRuntimeConnector.tsx
+++ b/frontend/src/components/molecules/DaRuntimeConnector.tsx
@@ -709,6 +709,20 @@ const DaRuntimeConnector = forwardRef<any, KitConnectProps>(
       }
     }
 
+    // Cleanup timeout on unmount
+    useEffect(() => {
+      return () => {
+        if (kitListTimeoutRef.current) {
+          clearTimeout(kitListTimeoutRef.current)
+          kitListTimeoutRef.current = null
+        }
+        if (socketConnectTimeoutRef.current) {
+          clearTimeout(socketConnectTimeoutRef.current)
+          socketConnectTimeoutRef.current = null
+        }
+      }
+    }, [])
+
     if (forceKitId) {
       let statusIcon = '🟡'
       let statusText = 'Connecting...'
@@ -754,20 +768,6 @@ const DaRuntimeConnector = forwardRef<any, KitConnectProps>(
         </div>
       )
     }
-
-    // Cleanup timeout on unmount
-    useEffect(() => {
-      return () => {
-        if (kitListTimeoutRef.current) {
-          clearTimeout(kitListTimeoutRef.current)
-          kitListTimeoutRef.current = null
-        }
-        if (socketConnectTimeoutRef.current) {
-          clearTimeout(socketConnectTimeoutRef.current)
-          socketConnectTimeoutRef.current = null
-        }
-      }
-    }, [])
 
     return (
       <div>
--- a/frontend/src/components/molecules/DaRuntimeConnector.tsx
+++ b/frontend/src/components/molecules/DaRuntimeConnector.tsx
@@ -369,7 +369,7 @@ const DaRuntimeConnector = forwardRef<any, KitConnectProps>(
       setSocketIo(io(kitServerUrl, effectiveSocketIoConfig))
       // Only re-create socket if kitServerUrl changes, not if config changes
       // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [kitServerUrl])
+    }, [kitServerUrl, JSON.stringify(socketIoConfig)])
 
     useEffect(() => {
       if (!socketio) return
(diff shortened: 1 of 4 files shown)


NhanLuongBGSV:
Hey @7inh, the feature direction is good but there are two blockers before this can merge:

**🔴 Blocker 1 — `useEffect` inside an `if` block (Rules of Hooks violation)**

In `DaRuntimeConnector.tsx`, there's a `useEffect` call inside the `if (forceKitId)` block:

```tsx
if (forceKitId) {
  // ...
  useEffect(() => {   // ← hooks cannot be called inside conditionals
    console.log(...)
  }, [...])

  return (...)
}
```

React's Rules of Hooks require hooks to always be called at the top level. This will crash in strict mode and cause unpredictable state corruption otherwise. The `useEffect` needs to be moved to the top level of the component.

**🔴 Blocker 2 — Typo in fallback URL (`PluginPageRender.tsx`)**

```tsx
// After this PR:
const KIT_SERVER_URL = runtimeServerUrl || config.runtime?.url || 'https://kit.digitalauto.teaaach'
```

`teaaach` should be `tech`. Any deployment without `RUNTIME_SERVER_URL` and without `config.runtime.url` falls back to a broken URL. Looks like a debug leftover.

**🟡 Please also clean up before merging:**
- 15+ `console.log`/`console.warn` calls are scattered throughout, including some in the render path (not inside effects) — these will fire on every render in production
- The socket `useEffect` dep array was changed from `[kitServerUrl, JSON.stringify(socketIoConfig)]` to `[kitServerUrl]` — if `effectiveSocketIoConfig` changes at runtime (e.g. site config update), the socket won't reconnect with the new config

PR #506, opened 2026-07-03
Fix feedback tab missing

7inh:
This pull request improves how tab configurations are managed and migrated in the frontend, with a focus on ensuring consistency and correctness when handling default and custom tabs. The main changes are the introduction of a utility to guarantee all default builtin tabs are present, updates to the migration logic, and refactoring of tab normalization in the template form.

### DIFF
--- a/frontend/src/components/molecules/PrototypeTabs.tsx
+++ b/frontend/src/components/molecules/PrototypeTabs.tsx
@@ -29,13 +29,31 @@ interface PrototypeTabsProps {
 
 
 // Default builtin tabs
-const DEFAULT_BUILTIN_TABS: TabConfig[] = [
+export const DEFAULT_BUILTIN_TABS: TabConfig[] = [
   { type: 'builtin', key: 'overview', label: 'Overview' },
   { type: 'builtin', key: 'journey', label: 'Customer Journey' },
   { type: 'builtin', key: 'code', label: 'SDV Code' },
   { type: 'builtin', key: 'dashboard', label: 'Dashboard' },
   { type: 'builtin', key: 'feedback', label: 'Feedback' },
 ]
+
+export function ensureDefaultBuiltinTabs(tabs: TabConfig[]): TabConfig[] {
+  const customTabs = tabs.filter((t) => t.type === 'custom')
+  const existingBuiltins = tabs.filter((t) => t.type === 'builtin')
+  const existingByKey = new Map(existingBuiltins.map((b) => [b.key, b]))
+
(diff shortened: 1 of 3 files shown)


NhanLuongBGSV:
Thanks for tackling this — the diagnosis is correct (old **customized** `prototype_tabs` records are frozen snapshots that don't include builtins added later, e.g. `feedback`), and reconcile-on-read is the right instinct. However, **we're going to hold this PR** — as-is it isn't a comprehensive solution, and it mixes a bugfix with a behavior change. Details below.

### 1. The backfill reshuffles tab order (behavior bug, not a nit)
`ensureDefaultBuiltinTabs` returns `[...mergedBuiltins, ...customTabs]`, which snaps builtins to canonical order and pushes **all custom tabs to the end**. So any stale record where the user reordered tabs — or placed a custom tab *between* builtins — gets silently rearranged the moment a builtin is backfilled. The reconcile should be **order-preserving**: only append/insert the genuinely-missing defaults and leave the existing layout untouched, e.g.

```ts
const missing = DEFAULT_BUILTIN_TABS.filter(d => !existingByKey.has(d.key))
return [...tabs, ...missing]   // or insert each missing default next to its canonical neighbor
```

### 2. It bundles a policy change with the bugfix
The `migrateTabConfig` backfill alone fixes the visible symptom (feedback renders again). The `TemplateForm` change — deleting `normalizeTabsForTemplate` in favor of `getTabConfig` — is separate: it removes the deliberate "a template may **omit** builtin tabs" contract (that helper's comment spelled this out), so templates can now only **hide** builtins, not exclude them. That's a product decision, not a bugfix, and shouldn't ride along in the same PR. Please split it out so it can be decided on its own.

### 3. Not comprehensive — the root cause is the data model
The underlying problem is that we persist a **full snapshot** of builtin tabs, so the stored array drifts out of sync with the live defaults forever. Reconcile-on-read patches the read path but leaves the DB stale and needs a matching patch at every read/write site. The comprehensive fix is to store only the **overlay/diff** (hidden flags, reordering, custom tabs) and always derive builtins from the current `DEFAULT_BUILTIN_TABS` — then a builtin can never go stale or be "lost," and there's nothing to reconcile. That's a larger data-model change and a separate ticket, but it's the direction we want.

### Suggested path forward
- Split this into: (a) an **order-preserving** read-time backfill (mergeable on its own), and (b) the template-semantics change as its own PR for a separate decision.
- Open a follow-up ticket for the overlay/diff data model as the real fix.

Appreciate the work — happy to review the split version.

PR #621, opened 2026-08-13
Add model visibility test

7inh:
Add model visibility test

### DIFF
--- a/.agents/TESTING.md
+++ b/.agents/TESTING.md
@@ -26,6 +26,7 @@ npx playwright test tests/prototype-context-menu.spec.ts
 npx playwright test tests/home-model-list.spec.ts
 npx playwright test tests/home-prototype-list.spec.ts
 npx playwright test tests/image-fallback.spec.ts
+npx playwright test tests/model-editable-visibility.spec.ts
 
 # With screenshots on failure
 npx playwright test --screenshot=only-on-failure
@@ -41,6 +42,7 @@ Playwright treats the file argument as a **regex**. Use **forward slashes** (not
 ```powershell
 cd C:\repo\autowrx\.agents
 npm run test:home-model-list
+npm run test:model-editable-visibility
 ```
 
 Or with `npx` (forward slashes work on Windows):
@@ -51,12 +53,14 @@ npx playwright test tests/import-export.spec.ts
 npx playwright test tests/prototype-context-menu.spec.ts
 npx playwright test tests/home-prototype-list.spec.ts
 npx playwright test tests/image-fallback.spec.ts
+npx playwright test tests/model-editable-visibility.spec.ts
 ```
 
 Run several suites:
 
 ```powershell
 npx playwright test tests/import-export.spec.ts tests/prototype-context-menu.spec.ts tests/home-model-list.spec.ts
+npx playwright test tests/model-editable-visibility.spec.ts
 ```
 
 ## Test Suites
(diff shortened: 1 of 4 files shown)


NhanLuongBGSV (changes requested):
## Review: Add model visibility test

Test-only PR (new E2E spec, 8 TCs, helper extensions, npm script, TESTING.md). I verified every endpoint, selector, and permission claim against the code on `main` — details below.

**Verdict: Request changes** — one process blocker, one state-leak bug.

### 🔴 Blocker: ECA check fails

`eclipsefdn/eca` is red — the author is not covered by the Eclipse legal agreements. Per `agentic/RULES.md`, this can't merge until the ECA is signed and the commit carries that identity. The PR body is also just the title; please fill in What / Why / How-verified per the repo's PR conventions.

### 🟠 TC7 permanently steals the default-template flag

TC7 creates a template with `is_default: true`. The backend's `create` then unsets **every** other default (`clearOtherDefaults`, `backend/src/services/modelTemplate.service.js:30-38`). The `afterEach` deletes the test template but never restores the previous default — a persistent/shared test environment is left with **no default model template at all**, silently changing behavior for later runs (e.g. new-prototype flows that preselect a default template).

Suggested fix: before creating, capture the current default template id (list templates with `is_default: true`) and restore its flag in cleanup — or create with `is_default: false` and assert the pairing via an update call instead, which also exercises the default-switching logic more precisely.

### ✅ What checks out (all verified on `main`, nothing fabricated)

- `editable` is a real visibility and publicly visible: `backend/src/config/enums.js:15`
- Non-owner create-on-model semantics (TC1/2/3): `canCreatePrototypeOnModel` passes non-owners only for `editable`, guests never — `backend/src/services/permission.service.js:288-308`
- Template → model visibility inheritance (TC8): `backend/src/services/model.service.js:56-57`; PATCH accepts `visibility` + `state` (`backend/src/validations/model.validation.js`)
- `/v2/system/model-template` POST (admin) / GET / DELETE (admin) all exist (`backend/src/routes/v2/system/modelTemplate.route.js`)
- Prototype create returns a bare id string (`backend/src/controllers/prototype.controller.js:22`) — the helper's defensive parsing is warranted
- Frontend selectors grounded: create-button wrapper classes `h-fit opacity-50 pointer-events-none` (`frontend/src/pages/PagePrototypeLibrary.tsx:152-156`); the new-prototype picker queries `{visibility: 'editable', state: 'released'}` (`frontend/src/components/molecules/forms/FormNewPrototype.tsx:116`) so TC6 is real; `readOnly: !editable` wiring for TC5 (`frontend/src/components/molecules/CodeEditor.tsx:113`)
- The shared `goToPrototypeCodeTab` relaxation (monaco-or-runtime-panel, 30s) does **not** weaken `prototype-runtime.spec.ts` — it awaits runtime readiness itself right after (`.agents/tests/prototype-runtime.spec.ts:35`)
- Hard `waitForTimeout` waits match existing suite style; all helper imports exist; license-header CI doesn't trigger here (workflow paths cover `backend/`/`frontend/` only, and sibling `.agents` specs carry no headers)

### 🟡 Nits

- TC5 "can edit" verifies editability only indirectly (API round-trip + absence of a monaco `readonly` class — that class assumption isn't verifiable from the repo). Acceptable pragmatic choice, but worth one confirmed green run.
- TESTING.md "Run several suites" section adds a standalone one-suite command instead of appending to the existing multi-file example.
- If `setModelReleasedViaApi` throws mid-helper, the already-created model id was never pushed to `createdModelIds` → orphan on the env. Same pattern exists in sibling suites, so consistent — noting only.

### ⚠️ Verification

I reviewed statically and did not execute the suite (it mutates models/users/templates on the configured environment). Since the PR pipeline runs the ECA check only, please run `npm run test:model-editable-visibility` locally and note the result in the PR description.

NhanLuongBGSV:
## Follow-up: suite executed on an isolated instance

**Result: 8/8 passed (TC1–TC8, ~2.8 min)** — run against a throwaway local stack (fresh DB, seeded admin), so "tests pass" is confirmed. As noted in the review, nothing in it predicted failures.

**TC7 default-template leak — now reproduced, not just code-read:**

- Pre-run: seeded one template, `is_default: true`
- Post-run: that template still exists but `is_default: false`; **no template in the environment has a default flag**
- The suite was green (8/8) while this happened — TC7's `POST` with `is_default: true` triggers `clearOtherDefaults()` (`modelTemplate.service.js:33-35`), `afterEach` deletes TC7's template, and the prior default is never restored

So local success and the leak are fully compatible: on a fresh dev stack there's often no prior default to lose, which is why it isn't visible to the author. Fix remains: capture the previous default before the test and restore it in cleanup (or assert the pairing via update instead of create).

The ECA check is still failing and is unaffected by test runs — it needs the contributor agreement signed for the commit identity.

7inh:
test merge main ok

Issue #79, opened 2025-04-29
Bug: In DaCustomPropertyItem, logic of function handleChange is wrong. If parameter name === "name", the value should be updated directly, but the code is currently checking for the type and transform the value


tuan-hda:
```js
} else {
    // Handle other changes like name or value based on the current type
    if (property.type === 'boolean') {
      updatedValue = value === 'true'
    } else if (property.type === 'number') {
      updatedValue = Number(value)
    } else if (property.type === 'null') {
      updatedValue = null
    }
    onUpdate({ ...property, [name]: updatedValue })
  }
```

This part modifies the value even if we are updating the 'name' field. This should be applied only when we update 'value' field.

PR #623, opened 2026-08-14
Align runtime `read-file`/`write-file` format and kit client registration  

7inh:
## Summary

Update the frontend kit protocol so Hardware Kit Manager and plugin pages emit `read-file` / `write-file` in the canonical `file_path` / `file_content` shape (with prototype and username metadata), parse both old and new reply shapes, and register browser clients with real display name + user id (no email). This matches the kit-side handlers landed in [sdv-runtime#81](https://github.com/eclipse-autowrx/sdv-runtime/pull/81).

## Motivation

- What problem does it solve? Load/Set config and VSS from Hardware Kit Manager (and plugin fetch/replace helpers) were emitting a legacy payload (`data` as path string or `{ path, content }`) that does not match the documented remote-access contract or the syncer handlers that implement it.
- What was difficult or missing before? Emit and reply shapes were inconsistent across `DaRuntimeConnector` and `PluginPageRender`; `register_client` used placeholders (`test` / `plugin`); reply parsing assumed a single nested `data` shape.
- Why is this approach useful? One wire format shared with sdv-runtime `#81`, dual reply parsing for transition hygiene, and safer client registration that avoids putting emails into Kit-Manager’s public client list.

## Changes

- Added/improved `read-file` / `write-file` emits in `DaRuntimeConnector` and `PluginPageRender` to use top-level `file_path` / `file_content`, plus `prototype` and `username`, with `data: ""`.
- Added `frontend/src/utils/kitReply.ts` helpers (`parseReadFileReply`, `prepareKitFileContentForWrite`) so replies support both flat (`result` / `has_error` / `file_path`) and nested (`data.path` / `data.content`) shapes, and writes normalize LF / strip trailing newlines.
- Updated Hardware Kit mode polling / subscription behavior in `DaRuntimeConnector` (`get-runtime-info` cadence, `unsubscribe_apis` on kit switch, richer `subscribe_apis` fields).
- Updated `register_client` to send signed-in **display name + user id only** (no email), with `'test'` / `'plugin'` fallbacks; re-emit when profile arrives after connect.
- Updated `docs/architecture/realtime-signals.md` with the payload/reply contract and registration privacy note.

## Behavior

Before:
- File commands used legacy `data` payloads; kits implementing the new contract did not receive `file_path` / `file_content`.
- Clients registered as placeholder `test` / `plugin`.
- Read replies that used the flat `result` / `has_error` shape were not reliably parsed.

After:
- Emit side speaks the new format expected by sdv-runtime `#81` (`/app/remote_access/...` paths).
- Reply parser accepts both old nested and new flat shapes.
- `register_client` announces real name/id without email (Kit-Manager exposes clients via public `GET /listAllClient`).

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- Code review against `sdv-runtime` `#81` handlers and `/app/remote_access` layout.
- Manual verification recommended: Hardware Kit Manager Load/Set for Config and VSS; Socket.IO frames show `file_path` / `file_content` on emit and `register_client` with name/id only; plugin fetch/replace signal mapping and VSS.

## Breaking Changes

Kits that only understand the legacy `data: { path, content }` shape and do **not** accept top-level `file_path` / `file_content` will no longer receive a usable write/read payload from this frontend. Kits on stock `sdv-runtime` main (no file handlers at all) already ignored both formats. Deploy [sdv-runtime#81](https://github.com/eclipse-autowrx/sdv-runtime/pull/81) (or an equivalent handler) on targeted kits.

## Additional Notes

- Companion PR: [eclipse-autowrx/sdv-runtime#81](https://github.com/eclipse-autowrx/sdv-runtime/pull/81) — syncer `read-file`/`write-file` + Dockerfile `/app/remote_access` (`vss.json` symlink to `/home/dev/ws/vss.json`, seeded `signal-config.json`).
- `read-file` / `write-file` `username` may still carry email for kit-side audit fields; that path is **not** exposed via `/listAllClient`. Only `register_client` was tightened for privacy.
- ECA commit signing remains a separate merge-gate item.
- Possible follow-ups: unit tests for `kitReply.ts`; optional name/id-only for file-command `username` as well; Kit-Manager auth on `/listAllClient`.


NhanLuongBGSV:
## Review: runtime read/write-file format update — verified against `sdv-runtime`

**What I checked:** cloned `eclipse-autowrx/sdv-runtime` (main) and compared every protocol change in this PR against the actual consumers — `Kit-Manager/src/index.js` (transparent relay) and `kuksa-syncer/syncer.py` (kit-side handler).

**✅ Aligned / improvements (vs sdv-runtime main):**
- `subscribe_apis` extra fields (`username`, `user_id`, `subscribed_at`) — kit reads only `apis`/`request_from`, extras ignored harmlessly; empty `apis: []` handled
- `get-runtime-info` 2s polling — implemented (`syncer.py:548`), negligible load
- `unsubscribe_apis` on kit switch — fixes a real stale-subscriber leak on kits
- Reply parser supporting both old (`data.path/content`) and new (`file_path`/`result`/`has_error`) shapes — good transition hygiene

**⚠️ Main open question — `read-file`/`write-file`:**
sdv-runtime main does **not implement these commands at all** (no handler, no `remote_access/` path — its paths are `/home/dev/ws`). So this format change can't be validated against that repo, and kits running it never answered these commands (old format included). Which runtime implements the new contract (`file_path`/`file_content`/`prototype`/`username`, `/app/remote_access/...`)? Please link its handler or the spec it was written against — and confirm it's deployed on kits targeted by the Hardware Kit Manager. Note the emit side is **new-format-only**; kits still expecting `data: {path, content}` would stop working.

**🔒 Privacy note:** `register_client` now sends real user emails to a broker with `cors: '*'` and a public, unauthenticated `GET /listAllClient` — previously placeholders leaked nothing. Consider sending name/id only, or hashing the email.

**Process:** ECA check is failing again — needs signing before merge.

NhanLuongBGSV:
## Follow-up review (1 week later) — still request-changes

State check: the PR is unchanged (head `98a64f54`) and the open questions from the initial review have had no response. I've also re-verified the ecosystem side:

**The implementing runtime still doesn't exist publicly.** `sdv-runtime` main is unchanged and still has no `read-file`/`write-file` handler (its syncer implements 14 commands — none are file commands; no `remote_access` paths). Its only open PR (#79, kit-manager diagnostics) doesn't add them either. So the headline change of this PR — the new `file_path`/`file_content`/`prototype`/`username` wire format — still cannot be validated against anything reachable, and per the original review the emit side is **new-format-only**.

**What that means in practice, restated:**
- If no runtime anywhere implements the new contract, the Hardware Kit Manager's read/write-config features would emit commands nothing answers (silent no-op/timeout after this merges)
- If any deployed kit still speaks the old `data: {path, content}` shape, this PR breaks file writes to it
- The `register_client` privacy concern also stands: real user emails to a broker with `cors: '*'` and public `GET /listAllClient` — name/id only, or hashed email, would be preferable

**To move this forward, either:**
1. Link the runtime/handler that implements the new file-command contract (repo + handler path), and confirm it's deployed on the kits the Hardware Kit Manager targets — happy to re-verify against it the same way I checked sdv-runtime, or
2. Park/close this until the runtime side actually lands — keeping a wire-format migration merged ahead of its only consumer risks exactly the breakage above with nothing gained

(ECA check remains failing, as with the rest of this batch.)

7inh:
### 1. `read-file` / `write-file` Consumer

The implementing runtime is now public: `eclipse-autowrx/sdv-runtime#81`
[https://github.com/eclipse-autowrx/sdv-runtime/pull/81](https://github.com/eclipse-autowrx/sdv-runtime/pull/81)

It includes:

* `kuksa-syncer/syncer.py` handlers.
* `/app/remote_access`.
* `vss.json → /home/dev/ws/vss.json` symlink.
* Seeded `signal-config.json`.

This matches this PR’s emit contract:

* `file_path` / `file_content`.
* Hyphenated commands.
* Dual reply shapes on the client.

Kits running that image are what Hardware Kit Manager targets for **Load/Set Config & VSS**.

**Note:** Stock `sdv-runtime` `main` never answered file commands in either format, so this PR is not a regression against that baseline. It aligns the frontend with the runtime that actually implements file I/O.

Legacy kits using `{path, content}` without the new handler would still not work. However, `#81`'s syncer also accepts the legacy shape if needed during the transition.

### 2. `register_client` Privacy

Agreed. Updated `register_client` to send only:

* Display name.
* User ID.

Email is no longer sent, since Kit-Manager stores clients and exposes them through unauthenticated `GET /listAllClient`.

Fallbacks remain `test` / `plugin`.

### 3. ECA

Separate merge-gate item. The commit will be signed before merge; this is orthogonal to the protocol review.

NhanLuongBGSV:
@7inh — the open question from the original review is now answered and verified. **Approve**, with one small residual.

### ✅ The implementing runtime exists and the contract matches

sdv-runtime#81 provides the `read-file`/`write-file` handlers and `/app/remote_access` provisioning. Cross-verified field-by-field against this PR's emit side: top-level `file_path`/`file_content` (preferred) with legacy `data` shapes still accepted, reply shape (`result`/`has_error`/`file_path`/`request_from`), fallback values (`no-name`/`no-id`/`no`), and the exact paths. The Aug-17 blocker — "new-format-only emits targeting nothing public" — is resolved.

### ✅ Backward compatibility confirmed

#81 parses **both** formats, so this PR can merge independently of #81's rollout: runtimes that never implemented file commands time out exactly as before (they always did), new runtimes answer. No breakage window.

### ✅ PII fix verified at head

`register_client` now sends `user?.name || 'test'` — no email — with a comment citing exactly why (the broker's public `GET /listAllClient`). The review's privacy note is properly closed for registration.

### 🟡 One residual: file-op `username` still carries the email

`read-file`/`write-file` payloads still send `currentUser?.email || currentUser?.name || 'no'` (`DaRuntimeConnector.tsx:393/405`, same in `PluginPageRender`). That value travels over the public broker into kit-side logs. Suggest mirroring the register fix (name only) — one-line change, and #81 could then log it as the audit identity it's presumably meant to be.

**Verdict: approve** — contingent on nothing; merge when ready (ECA aside). The companion runtime PR sdv-runtime#81 has one security item of its own (path confinement) noted on its thread — it does not gate this PR.

7inh:
Mirrored the register fix: read-file/write-file username is now display name only ('no' fallback), so kit-side audit logs get a stable identity without email over the public broker.

tri2510:
@7inh — residual verified fixed at `5d1bc5f9`: **final approve, merge-ready.**

All emit sites confirmed name-only at head — `DaRuntimeConnector` (register, subscribe, read/write-file) and `PluginPageRender` (register + four file-op sites) now use `currentUser?.name` with the `'no'`/`'test'`/`'plugin'`/`'anonymous'` fallbacks. Zero email reaches the broker; the remaining `email` references in the file are the user-search feature and React dependency-array leftovers (inert — they only list a value the callbacks no longer read).

That closes every item across the review history: the implementing runtime identified and contract-matched (#81), backward compatibility confirmed (both formats accepted), registration PII fixed, and file-op usernames now a stable display-name identity for kit-side audit logs — which #81's new audit logging consumes.

**Verdict: approve.** The pair (#623 here + sdv-runtime#81) can merge in either order.

PR #644, opened 2026-08-17
fix: serialize populated model ids in has-permission query

7inh:
# Summary

Fix permission query serialization so a populated model object is sent as a real ID (`writeModel:<id>`) instead of `writeModel:[object Object]`. Prototype detail pages can then resolve `WRITE_MODEL` correctly.

# Motivation

On the prototype detail page, `useCanEditPrototype` calls:

```ts
usePermissionHook([MANAGE_USERS], [WRITE_MODEL, prototype.model_id])
```

`GET /v2/prototypes/:id` populates `model_id` (`getPrototypeById` uses `populate({ path: 'model_id' })`), so the frontend receives `{ id, name, visibility }` rather than a string.

`checkPermissionService` used `perm.join(':')`, which stringified that object to `[object Object]`.

The request looked like:

```text
/v2/permissions/has-permission?permissions=manageUsers,writeModel:%5Bobject+Object%5D
```

`[object Object]` is not a valid `ObjectId`. Mongoose throws `CastError` (400). The controller `Promise.all` then fails the entire batch, so `manageUsers` is also lost.

`usePermissionHook` falls back to `fill(false)` — fail-closed, but edit/admin UI on that page was wrong.

## Why this approach?

One change in `checkPermissionService` covers every caller.

Empty/null refs still become a bare action (`writeModel`), which matches the old `writeModel` global-check path.

No backend populate change is needed; that populate is still required for visibility checks.

# Changes

* Added `toPermissionRef()` in `frontend/src/services/permission.service.ts` to:

  * Take a string as-is.
  * Extract `id` from a populated object.
  * Fall back to `_id` from a populated object.
* Updated `checkPermissionService` to build `ACTION:ref` from that helper instead of using `Array.join(':')`.

# Behavior

## Before

Prototype detail pages (and any hook passing a populated `model_id`) requested:

```text
writeModel:[object Object]
```

The batch returned 400, causing both write and admin flags on that page to become `false`.

## After

The same call becomes:

```text
writeModel:6a7e8a1690eecd2e4e9e523e
```

Scoped `WRITE_MODEL` and sibling `manageUsers` permissions in the same request both return real booleans.

Missing refs are still sent as the bare action name.

# Testing

* [ ] Unit tests
* [ ] Integration tests
* [x] Manual testing

**Tested with:**

Open a prototype detail page and confirm the network call is:

```text
/v2/permissions/has-permission?permissions=manageUsers,writeModel:<actualModelId>
```

and not:

```text
writeModel:%5Bobject+Object%5D
```

Check that contributor write UI and admin actions match the user's actual permissions.

# Breaking Changes

None.

# Additional Notes

* **Trigger:** `GET /v2/prototypes/:id` → populated `model_id` → `useCanEditPrototype` (`frontend/src/hooks/useCanEditPrototype.ts`).
* An object with neither `id` nor `_id` is sent as a bare action (global check). This is true only for users who already have the permission on `*` (e.g. admin `WRITE_MODEL`). It does not grant write access to a normal contributor. Writes remain enforced server-side.
* **Optional follow-up:** Extract a string ID in `useCanEditPrototype` before calling the hook. This is not required for the fix.

### DIFF
--- a/frontend/src/services/permission.service.ts
+++ b/frontend/src/services/permission.service.ts
@@ -8,14 +8,29 @@
 
 import { serverAxios } from './base'
 
+const toPermissionRef = (value: unknown): string | undefined => {
+  if (value == null || value === '') return undefined
+  if (typeof value === 'string') return value
+  if (typeof value === 'object') {
+    const obj = value as { id?: unknown; _id?: unknown }
+    if (typeof obj.id === 'string' && obj.id) return obj.id
+    if (obj._id != null) return String(obj._id)
+  }
+  return undefined
+}
+
 export const checkPermissionService = async (
   permissions: [string, string?][],
 ) => {
+  const list = Array.isArray(permissions) ? permissions : [permissions]
   return (
     await serverAxios.get<boolean[]>(`/permissions/has-permission`, {
       params: {
-        permissions: (Array.isArray(permissions) ? permissions : [permissions])
-          .map((perm) => perm.join(':'))
+        permissions: list
+          .map(([action, ref]) => {
+            const refId = toPermissionRef(ref)
+            return refId ? `${action}:${refId}` : action
+          })
           .join(','),
       },
     })


NhanLuongBGSV:
## Review: serialize populated model ids in has-permission query

**What it is:** one-file fix in `permission.service.ts` — `checkPermissionService` builds `ACTION:ref` strings via `perm.join(':')`, so a populated object ref (e.g. `prototype.model_id` passed straight through at `useCanEditPrototype.ts:23`) produced `WRITE_MODEL:[object Object]`. The new `toPermissionRef()` extracts `id`/`_id` from objects.

**Verdict: approve.** Verified mechanics against the backend:

- **Old failure mode was safe-deny:** `WRITE_MODEL:[object Object]` → backend `Model.findById('[object Object]')` misses → 404 thrown → request rejects → `usePermissionHook` returns `Array(params.length).fill(false)` (`usePermissionHook.ts:27`) → UI denied. Fail-closed, but broken.
- **The fix is correct for the real case:** object-with-id now yields the proper scoped `WRITE_MODEL:<id>` check — strictly better.
- **Empty/null refs are unchanged:** the old `join(':')` on `[action, undefined]` produced `"WRITE_MODEL:"` → backend `split(':')` gives `id: ''` → falsy → global check (`permission.service.js:196-198`). The new bare `WRITE_MODEL` hits the same global-check path. No semantic change there.
- **Residual exposure is negligible:** an object carrying neither `id` nor `_id` would now get the global check instead of the old 404-deny — but populated docs always carry one of those, and every write is enforced server-side regardless, so this can only affect UI control visibility, not data.

### Suggestions (non-blocking)

- When the ref is an unresolvable object, omitting the pair entirely (hook fills `false`) would preserve fail-closed semantics exactly, instead of falling back to the global check. Fine as-is given the analysis above.
- Naming the endpoint that actually returns a populated `model_id` in the PR description would help future readers — grep of the prototype controller/service shows no `populate('model_id')`, so the trigger may be defensive rather than live today.

No tests (frontend runner gap, known). License header intact, CI green.

**Process:** ECA check failing — flagged; merge is a maintainer policy call.

7inh:
Agree on the approval and on empty-ref parity. Two corrections:

1. The old path was a **Mongoose `CastError` → 400** on the whole `Promise.all`, not `findById('[object Object]') → 404`. That also zeroed `manageUsers` in the same batch.
2. The populate is live: `GET /v2/prototypes/:id` (`getPrototypeById` uses `populate({ path: 'model_id' })`), which is why `useCanEditPrototype` saw an object. I’ll mention that endpoint in the PR description.

Omitting an unresolvable pair would desync the result array from the hook params. Leaving it as bare `writeModel` is enough — that only passes for `*` holders who already pass the scoped check.

NhanLuongBGSV:
**Live evidence for this fix** — captured from production traffic on a freshly deployed instance (v3stage, v2026.08.18), seconds after deploy:

```
GET /v2/permissions/has-permission?permissions=manageUsers,writeModel:%5Bobject+Object%5D
500 — Cast to ObjectId failed for value "[object Object]" (type string) at path "_id" for model "Model"
```

A real user session on a prototype page sent `writeModel:[object Object]` — the populated-ref case this PR fixes — producing a live HTTP 500 (permission check then fails closed on the client). So the trigger is active production behavior, not defensive: some prototype-detail payloads carry a populated `model_id` that flows straight into `useCanEditPrototype`.

This also resolves the open question from review: *"which endpoint actually returns a populated `model_id`"* — it occurs on prototype detail views. With this PR, that request becomes a proper scoped `writeModel:<id>` check instead of a 500.

7inh:
Makes total sense. Thanks for breaking down the errorConverter logic and identifying GET /v2/prototypes/:id as the source API. Good catch on the distinction between the page view and the underlying endpoint!

NhanLuongBGSV:
## Independent verification — APPROVE

A second independent review pass confirmed the analysis end-to-end. Summary for the record:

**✅ Bug chain confirmed with citations:** `getPrototypeById` populates `model_id` (`prototype.service.js:171-178`) → decorator passes it through (`FeedbackPrototypeDecorator.js:13-20`) → the toJSON plugin guarantees the populated doc carries a string `id` (`toJSON.plugin.js:39-40`), so `toPermissionRef`'s `id`-first ordering matches this codebase's serialization convention → `useCanEditPrototype.ts:20-23` passes the object verbatim → old `join(':')` produced `writeModel:[object Object]`. Seven components consume `useCanEditPrototype`; all were affected.

**✅ Status-code question settled (400 vs 500 — both correct, different layers):** `errorConverter` maps the Mongoose `CastError` to **400** (`middlewares/error.js:15-25`), but marks the converted ApiError non-operational; `errorHandler` then overrides non-operational errors to **500 in production** (`error.js:29-33`). So dev = 400, prod = 500 — which is exactly why the v3stage production log showed 500. (My earlier "404" was wrong; apologies for the churn.)

**✅ Empty/null-ref path provably unchanged:** old `['writeModel', undefined].join(':')` → `"writeModel:"` → backend `split(':')` → `id: ''` (falsy) → global check; new bare `writeModel` → `query[1]` undefined (falsy) → same global check. Identical semantics.

**✅ Fallback risk is display-only:** a bare action can only match a wildcard (`roleMap.has('*')`) grant (`permission.service.js:146-159`) — i.e., admins who already see those controls; every write is independently enforced server-side (`auth() + checkPermission` on prototype/model/asset routes).

**🟡 Root-cause follow-up (pre-existing, out of scope here):** `Prototype.model_id` is typed `string` (`frontend/src/types/model.type.ts:82`) but detail responses return an object. Other consumers still stringify it: `PrototypeTabInfo.tsx:130-133` and `PrototypeTabJourney.tsx:64-67` post `parent_id: model_id` (an object) into activity logs, producing `[object Object]` in log descriptions. A follow-up typing/normalization fix would close the whole class.

**Verdict: approve** — correct, minimal, fails safe. (ECA check remains the standing policy gate.)

PR #645, opened 2026-08-17
fix: cache and dedupe public site-config fetches

7inh:
## Summary

Share one cached bulk `GET /site-config/public` request across all `useSiteConfig` / `getConfig` callers so the prototype library no longer fires a separate `DEFAULT_PROTOTYPE_IMAGE` request per card.

## Motivation

**What problem does it solve?**

Opening Prototype Library (and similar lists) caused many identical calls to:

`GET /v2/site-config/public/DEFAULT_PROTOTYPE_IMAGE?scope=site`

Each `DaPrototypeCard` called `useDefaultPrototypeImage()` → `useSiteConfig()`, and that hook always hit the per-key public endpoint with no shared cache.

**What was difficult or missing before?**

* No in-flight dedupe: 20 cards mounting together meant 20 identical HTTP requests.
* No shared single-key cache: remounts/re-renders fetched again.
* Bulk `/public` was already cached, but single-key lookups did not use it.
* Nav/layout also fetched other keys (`SITE_TITLE`, `SITE_LOGO_WIDE`, …) as separate per-key requests.

**Why this approach?**

Keep the existing `useSiteConfig` / `getConfig` API so cards and lists do not need to change. Make `getConfig` read from one bulk public-config fetch (cached 5 minutes, in-flight requests shared). All keys in the same scope reuse that one response.

## Changes

* Updated `frontend/src/utils/siteConfig.ts` to use one bulk public-config cache as the source of truth for `getConfig`, `getConfigs`, `getPublicConfigs`, and `useSiteConfig`.
* Added in-flight dedupe so concurrent callers share the same request.
* Per-scope cache expiry (no longer one global expiry for all scopes).
* Fixed stale sync helpers (`getSiteConfigSync`, `getConfigsSync`) so they read the real bulk cache key.
* `useSiteConfig` / `useSiteConfigs` ignore results after unmount.
* No component changes in `DaPrototypeCard` or `PrototypeLibraryList`.

## Behavior

**Before:**

* N prototype cards → N calls to `/site-config/public/DEFAULT_PROTOTYPE_IMAGE`.
* Nav/layout each unique key → its own `/site-config/public/{KEY}` request.

**After:**

* One `GET /v2/site-config/public?scope=site` per scope for 5 minutes.
* Cards, nav, and layout read keys from that cached object.
* Missing/null values still fall back to the hook argument or `DEFAULT_SITE_CONFIGS`.

## Testing

* [ ] Unit tests
* [ ] Integration tests
* [x] Manual testing
* [ ] Existing test suite passes (frontend has no unit-test runner)

**Tested with:**

* Open Prototype Library with many cards.
* Network tab filtered to `site-config/public`.
* Expected: **one** bulk `/public?scope=site` request, **zero** `/public/DEFAULT_PROTOTYPE_IMAGE` from cards.
* Also check home prototype/model lists and nav logo/title still resolve correctly.

## Breaking Changes

None. Public helpers (`getConfig`, `useSiteConfig`, `useDefaultPrototypeImage`, `useDefaultModelImage`) keep the same signatures.

## Additional Notes

* Direct `configManagementService.getPublicConfig(...)` callers are unchanged (home `CFG_HOME_CONTENT`, staging, privacy policy). Those still use the per-key route.
* `useAuthConfigs` still fetches bulk public configs on its own; wiring it through `getPublicConfigs` would collapse a duplicate bulk call (follow-up).
* `clearCache()` is unused today. If we later call it after admin save, an in-flight request can write pre-clear data back into the cache; fix then with a generation token or skip-write-after-clear.
* Cache entries expire lazily and are not pruned. Current callers are site-scope only, so this is one Map entry in practice.
* Backend `GET /site-config/public/:key` still does not filter `secret: false` (pre-existing). Out of scope; worth a separate backend PR on both per-key public routes.

### DIFF
--- a/frontend/src/utils/siteConfig.ts
+++ b/frontend/src/utils/siteConfig.ts
@@ -270,7 +270,9 @@ export const useSiteConfigs = (
         }
       } catch (err) {
         if (!cancelled) {
-          setError(err instanceof Error ? err.message : 'Failed to fetch configs')
+          setError(
+            err instanceof Error ? err.message : 'Failed to fetch configs',
+          )
         }
       } finally {
         if (!cancelled) {
@@ -301,7 +303,10 @@ export const useSiteConfig = (
 ) => {
   const defaultVal = fallback ?? DEFAULT_SITE_CONFIGS[key]
   const [value, setValue] = useState<any>(() =>
-    parseConfigValue(getConfigSync(key, defaultVal, scope, target_id), defaultVal),
+    parseConfigValue(
+      getConfigSync(key, defaultVal, scope, target_id),
+      defaultVal,
+    ),
   )
 
   useEffect(() => {
--- a/frontend/src/utils/siteConfig.ts
+++ b/frontend/src/utils/siteConfig.ts
@@ -9,10 +9,57 @@
 import { useState, useEffect } from 'react'
 import { configManagementService } from '../services/configManagement.service'
 
-// Cache for site configs to avoid repeated API calls
-let configCache = new Map<string, any>()
-let cacheExpiry: number | null = null
 const CACHE_DURATION = 5 * 60 * 1000 // 5 minutes
+
+const DEFAULT_SITE_CONFIGS: Record<string, any> = {
+  SITE_LOGO_WIDE: '/imgs/logo-wide.png',
+  DEFAULT_MODEL_IMAGE: '/imgs/default-model-image.png',
+  DEFAULT_PROTOTYPE_IMAGE: '/imgs/default_prototype_cover.jpg',
+  SITE_TITLE: 'AutoWRX',
+  SITE_DESCRIPTION: 'Vehicle Signal Specification Management Platform',
+  SITE_FAVICON: '/imgs/favicon.ico',
+  SITE_THEME_COLOR: '#198100',
+  DISABLE_CUSTOM_API_SETS: false,
+  GENAI_SDV_APP_ENDPOINT:
+    'https://workflow.digital.auto/webhook/c0ba14bc-c6a3-4319-ad0a-ad89b1460b36',
+  VSS_PLUGINS: [{ label: 'A2L Importer', plugin: 'a2l-importer' }],
+}
+
+type CachedBulkConfigs = { data: Record<string, any>; expiry: number }
+
+const bulkCache = new Map<string, CachedBulkConfigs>()
+const inFlightBulk = new Map<string, Promise<Record<string, any>>>()
+
+const makeBulkCacheKey = (scope: string, target_id?: string) =>
+  `public_${scope}_${target_id ?? ''}`
+
+const readCachedBulk = (
+  scope: string,
+  target_id?: string,
+): Record<string, any> | null => {
+  const cached = bulkCache.get(makeBulkCacheKey(scope, target_id))
+  if (cached && Date.now() < cached.expiry) {
+    return cached.data
+  }
+  return null
+}
+
+const resolveConfigValue = (raw: any, key: string, defaultValue: any): any => {
+  if (raw !== null && raw !== undefined) {
+    return raw
+  }
+  return defaultValue !== null ? defaultValue : DEFAULT_SITE_CONFIGS[key]
+}
+
+const pickConfigValue = (
+  configs: Record<string, any> | null | undefined,
+  key: string,
+): any => {
+  if (configs && Object.prototype.hasOwnProperty.call(configs, key)) {
+    return configs[key]
+  }
+  return undefined
+}
 
 const parseConfigValue = (raw: any, defaultVal: any): any => {
   if (raw === null || raw === undefined || raw === '') {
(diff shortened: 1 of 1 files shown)


NhanLuongBGSV:
## Review: cache + dedupe public site-config fetches

**What it is:** refactor of `frontend/src/utils/siteConfig.ts` — the old cache (one Map + one *global* expiry shared across all scopes, per-key requests from `getConfig`) becomes a per-scope/target bulk cache with individual expiries, in-flight promise deduplication, and `getConfig`/hooks reading from one shared `/site-config/public` fetch.

**Verdict: approve with nits** — this fixes the stated problem and several latent bugs verified against `main`.

### ✅ Real fixes

1. **The N-requests problem, properly solved.** 50 prototype cards mounting at once now share one in-flight promise (`inFlightBulk` set before `await`, cleared in `finally` — correct pattern), and subsequent mounts read the warm cache synchronously in `useSiteConfig`'s early-return path with zero network.
2. **Sync cache reads never worked before — they do now.** Old `getConfigsSync` read cache key `'public'`, but the writer stored `public_${scope}_${target}` → always missed → always refetched. Old `getSiteConfigSync` read `public_site_` while the writer stored `public_site_site` → also always missed. The unified `makeBulkCacheKey` fixes both.
3. **Per-entry expiry** replaces the global `cacheExpiry` — previously one scope's refresh kept other scopes' stale entries "fresh". Each `(scope, target_id)` now expires independently.
4. **Unmount safety:** both hooks gain `cancelled` flags (no setState-after-unmount); `Object.is` functional updates avoid redundant re-renders.
5. **Incidental security improvement:** old `getConfig` called `GET /site-config/public/:key`, which resolves via `getSiteConfigByKey` — query `{key, scope}` with **no `secret` filter** (`backend/src/services/siteConfig.service.js:504-510`). The "public" per-key route could return secret-marked config values. The new path uses `GET /site-config/public` → `getPublicSiteConfigs`, which filters `secret: false` (`:603-604`). The frontend no longer exercises the leaky route.

### 🟡 Follow-ups / nits

1. **The backend hole is still open:** `GET /site-config/public/:key` (`backend/src/routes/v2/system/site-management.route.js:22`) remains unfiltered server-side — it can be called directly regardless of this frontend change. Worth a separate small backend PR adding the same `secret: false` filter (and the scoped variant `/public/:scope/:target_id/:key` likely too).
2. `clearCache()` during an in-flight fetch lets the completing request re-insert pre-clear data for another 5 min — same race class as before, not a regression, just worth knowing.
3. `bulkCache` entries expire lazily and are never evicted — model-scoped `target_id`s accumulate. Trivial footprint; an opportunistic prune would tidy it.
4. No tests — the frontend has no runner (known repo gap); the new pure helpers (`makeBulkCacheKey`, `pickConfigValue`, `resolveConfigValue`, `readCachedBulk`) are first candidates if one lands.
5. Minor: the per-key route's inline-default fallback has no bulk equivalent — values stored as `null` now fall through to `DEFAULT_SITE_CONFIGS` instead; those mirror the same inline defaults, so equivalent in practice.

**Process:** ECA check failing — needs signing before merge.

7inh:
> Backend GET /public/:key is unfiltered 

— correct, but out of scope

> clearCache() during an in-flight fetch lets the completing request re-insert pre-clear data for another 5 min

 —  clearCache is defined and never called. Nothing currently says “admin saved a config, wipe the
  cache.” Until that exists, this race cannot happen in production. If you later call clearCache() after an admin save, then it matters.

> bulkCache entries expire lazily and are never evicted 

 — technically true, practically empty

> No tests — the frontend has no runner 

— correct as a repo fact, not a PR gap

>  Null vs inline default

— In practice it is still a wash for this PR: staging still uses the per-key API; most hooks pass their own fallback; predefined configs are seeded so values are rarely null. Not a behavior change you should block on.

NhanLuongBGSV:
## Production-safety verification — SAFE TO MERGE

An independent deep review was run to catch production breakage before merge (live-DB cross-reference, full consumer sweep, old-vs-new semantics diff, and an actual `tsc + vite build` on this branch). Verdict: **no production-breaking defect**. Findings:

### ✅ Cleared with live evidence

1. **No silent value loss from the endpoint switch** (per-key `/public/:key` → bulk `/public`, which filters `secret: false`). Queried the live config DB (read-only): exactly **one** secret key exists — `EMAIL_CONFIG` — and it is *not* consumed through this module (the email admin UI uses the authenticated admin endpoint). All 22 keys consumed via `getConfig`/`useSiteConfig` are `secret: false`. The two consumed keys absent from the DB (`VSS_PLUGINS`, `HOME_HERO_IMAGE`) fall to identical frontend defaults in old and new code.
2. **No flipped feature flags.** Every boolean key's stored value was compared with its frontend fallback — identical end states everywhere. The warm-cache first paint actually *improves* `SHOW_CODE_DIFF` and `SHOW_SDV_PROTOPILOT_BUTTON`: the correct value renders immediately where the old code flashed the fallback for a frame.

### ✅ Consumer sweep & build

- Exhaustive grep: the only externally consumed exports are `getConfig`, `useSiteConfig`, and the two image wrappers (34 importing files, all site-scope) — the remaining exports have zero external callers, so their behavior changes are unobservable. All 11 exports keep identical signatures.
- `npm run build` (tsc + vite): **PASS** on this branch. (`npm run lint` fails repo-wide on a pre-existing eslint peer conflict — identical on base, not from this PR.)
- In-flight dedup races: none (the shared promise never rejects; `finally` cleanup is correct). Memory: one ~19KB cache entry in practice, replacing N per-card requests for the same already-public data. Hook cleanup/dep arrays unchanged-or-better; StrictMode-safe.

### 🟡 One bounded, observable change

Already-open SPA sessions can see an admin-saved config up to **5 minutes late** (old code fetched fresh on every mount; the new 5-min TTL caches). Hard reload clears instantly; E2E specs are unaffected (they verify after full page loads). Follow-up suggestion: wire the existing `clearCache()` into the admin config-save flow.

### 🔧 Small pre-merge nit (introduced by this PR)

`npx prettier --check src/utils/siteConfig.ts` fails on two long lines (~273, ~304) — the husky pre-commit only runs `tsc`, so it slipped through. A quick `prettier --write src/utils/siteConfig.ts` commit would keep CI formatting clean. No runtime impact.

Also noted: the old per-key endpoint silently ignored `scope`/`target_id`; the new bulk path honors them — strictly more correct for future callers.

7inh:
> 🟡 One bounded, observable change

Correct. The 5-minute delay is an acceptable trade-off for this PR given the reduction in redundant requests. We can add cache invalidation after config saves as a separate follow-up, since changing refresh behavior here would expand scope and could affect existing config-update flows.

7inh:
Test merged main good

PR #646, opened 2026-08-18
Feat/builtin genai forwarder

7inh:
# Summary

Replace the `genai-proxy` sidecar with a built-in backend forwarder, and add an authenticated `GET /v2/genai/token` issuer so plugins can mint short-lived Azure Speech credentials without exposing platform secrets.

Speech token minting reads platform environment variables (`AZURE_SPEECH_SDK_KEY` / `AZURE_SPEECH_SDK_REGION`), bypasses the corporate HTTP proxy for internal/STS calls, and falls back to the subscription key when Azure STS is unreachable.

# Motivation

* GenAI plugin traffic depended on a sidecar (`GENAI_URL` → `genai-proxy`). Deployments that already run GenAI as `http://genai:8080` still had to operate a second hop.
* Voice input needed a speech token. The host `api.getSpeechToken` was not always available, and documenting Azure-specific environment names did not match the platform-injected `AZURE_SPEECH_SDK_*` variables.
* `GET /v2/genai/actuator/info` returned `503 No response received from server` with no useful production log because `logger.debug` is silent when `NODE_ENV=production`. Axios also honored `HTTP_PROXY`, so internal `http://genai:8080` and STS requests could be sent through `rb-proxy-apac.bosch.com`.
* When STS (`…/sts/v1.0/issueToken`) is blocked by egress, voice failed even though the subscription key was present on the pod.

* No first-class `/v2/genai/token` route; speech depended on a host helper.
* Per-environment URLs (`EXTERNAL_GENAI_URL_DEV/PROD/STAGING`) added operator surface without matching this instance, which has one upstream.
* Failures were opaque in production logs.

## Why is this approach useful?

* One URL + device token is enough to forward GenAI requests.
* Token issuance is JWT-gated, keeps long-lived secrets on the server when STS works, and still enables voice when STS cannot be reached.
* Info/warn logs make the next `503` diagnosable with the method, path, upstream URL, and Axios error code.

# Changes

* Added a built-in GenAI forwarder:

  * `externalForward.service.js`
  * `genai.service.js`
  * Handles `/v2/genai/*`
  * JWT `auth()`
  * JSON + SSE
  * Profile VSS fallback
* Added `GET /v2/genai/token?service=azure-speech` with an issuer map, currently Azure Speech first.
* Updated configuration to use:

  * `EXTERNAL_GENAI_URL`
  * `EXTERNAL_GENAI_DEVICE_TOKEN`
* Removed:

  * `EXTERNAL_GENAI_URL_DEV`
  * `EXTERNAL_GENAI_URL_PROD`
  * `EXTERNAL_GENAI_URL_STAGING`
* Updated the speech issuer to read:

  * `AZURE_SPEECH_SDK_KEY`
  * `AZURE_SPEECH_SDK_REGION`
  * Fallback: `AZURE_SPEECH_KEY` / `AZURE_SPEECH_REGION`
* Refactored GenAI Axios calls to use `proxy: false` so Docker/internal and STS requests are not sent through `HTTP_PROXY`.
* Added production-visible `[genai]` / `[serviceToken]` info/warn logs without exposing secrets.
* Added STS failure fallback:

  * Returns the original subscription key.
  * Sets `auth: "subscription"` instead of returning `502`.
* Updated documentation and samples:

  * `backend/docs/genai-service.md`
  * `README`
  * `.env.example`
  * `instance-setup/.env.prod.sample`
* Added tests for the service-token issuer:

  * Unknown service
  * Missing credentials
  * STS success
  * STS fallback

> **Before merge:** confirm these tests are actually committed to the branch; they may still be local-only.

# Behavior

## Before

* `/v2/genai/*` required the `genai-proxy` sidecar (`GENAI_URL`) or returned `not implemented`.
* Voice depended on `api.getSpeechToken` from the host.
* Optional per-environment GenAI base URLs were supported.
* Production `503`s on `/v2/genai/actuator/info` had no `[genai]` log line.
* If STS was blocked, `/v2/genai/token` failed even when `AZURE_SPEECH_SDK_*` was set.
* Previously, token issuance also failed with `not configured` when only the SDK-specific environment variable names were available.

## After

* With `EXTERNAL_GENAI_URL` + `EXTERNAL_GENAI_DEVICE_TOKEN`, the backend calls GenAI directly using `Authorization: Token …`.
* `GENAI_URL` remains a deprecated fallback only when the new variables are unset.
* `GET /v2/genai/token?service=azure-speech` requires JWT authentication and a cookie, and returns:

```json
{
  "token": "...",
  "region": "...",
  "auth": "authorization"
}
```

* When STS succeeds:

  * `auth: "authorization"`
  * Returns a short-lived Azure Speech token.
* When STS is unreachable:

  * Returns HTTP `200`.
  * `auth: "subscription"`
  * Returns the original subscription key.
* Clients must choose between `Speech SDK fromAuthorizationToken` and `fromSubscription` based on the `auth` field.
* Voice can be gated on a successful token fetch on the plugin side.
* Production logs include the inbound path, upstream URL, and Axios error code without logging keys or tokens.
* Path suffixes `/dev`, `/prod`, and `/staging` are still stripped to `/generation` or `/profiles/{id}` on the single configured URL.

# Testing

* [x] Unit tests — service-token issuer (`400` / `503` / STS success / STS → subscription fallback)
* [ ] Integration tests
* [x] Manual testing
* [ ] Existing test suite passes — run `yarn test` in `backend/` before merge

**Tested with:**

* CloudHub instance: `GET /v2/system/plugin/slug/genai` → `200`
* `GET /v2/genai/token?service=azure-speech` with `AZURE_SPEECH_SDK_*` on the pod after the SDK environment-name fix.
* Production logs showing:

  * `[genai] GET /v2/genai/token?service=azure-speech configured=true`
  * `[serviceToken] Issuing token…`
* Local Jest:

  * `tests/unit/services/serviceToken.service.test.js`

# Breaking Changes

None for existing JWT GenAI plugin clients using:

* `/v2/genai/generation`
* `/v2/genai/profiles/{id}`

## Non-breaking Notes

* Prefer `EXTERNAL_GENAI_URL` + `EXTERNAL_GENAI_DEVICE_TOKEN`; stop relying on `genai-proxy` / `GENAI_URL`.
* Per-environment URL variables are ignored/removed; configure one URL.
* Token JSON now includes `auth`. Older clients that always call `fromAuthorizationToken` will break if they receive `auth: "subscription"`. Pair this change with the `genai-wizard` plugin change.

# Additional Notes

* Do not log device tokens, speech keys, or `Authorization` headers.
* Speech secrets are not included in Joi/README configuration; they come from platform environment variables (`AZURE_SPEECH_SDK_*`).
* `proxy: false` is required on Calponia when `HTTP_PROXY` is set. Otherwise, internal `genai:8080` and STS requests go through the Bosch proxy and can fail with `ENOTFOUND rb-proxy-apac.bosch.com`.
* Falling back to the subscription key exposes the long-lived key to the authenticated browser client. This is intentional so voice continues to work without STS egress.
* Prefer STS when egress to `*.api.cognitive.microsoft.com` is allowed (`com.calponia.egress.fqdns` on `playground-be`).
* Browser Speech SDK still needs network access to Azure for recognition; the fallback only skips the backend → STS hop.
* Follow-ups:

  * Commit/include `backend/tests/unit/services/serviceToken.service.test.js` if missing.
  * Add Calponia egress for `cognitive.microsoft.com` on `playground-be` if STS-based authentication is required.
  * Plugin-side auth handling is in `plugins/genai-wizard`.


NhanLuongBGSV:
## Review: builtin GenAI forwarder + service token issuer

**What it is:** replaces the `genai-proxy` sidecar with a built-in backend forwarder (`EXTERNAL_GENAI_URL` + device token), and adds authenticated `GET /v2/genai/token` so plugins can mint short-lived Azure Speech credentials without embedding platform secrets.

### 🔴 Blocker: the fallback leaks the platform subscription key

In `serviceToken.service.js`, when Azure STS is unreachable (or returns an empty body), `issueAzureSpeechToken` **returns the raw `AZURE_SPEECH_SDK_KEY` to the client**:

```js
const subscriptionResult = { service, token: azureKey, region, auth: 'subscription' };
...
} catch (error) { ... return subscriptionResult; }
```

Any authenticated user who triggers the fallback — STS hiccup, egress blocked, or simply an empty response — receives the long-lived, billable platform key in the response body. That defeats the PR's own stated goal ("without exposing platform secrets"). Please return **503** on STS failure and let the client show "speech unavailable" — never fall back to the key.

### 🟠 Should decide: minting has no RBAC or rate limit

`GET /v2/genai/token` sits behind plain `auth()` — any logged-in user (including a brand-new account) can mint unlimited STS tokens (each valid ~10 min, billable). Consider a permission check and/or simple rate limiting per user.

### ✅ What's well done

- Forwarder design: env-fixed target (no SSRF), device-token auth only — the user's credentials are **not** forwarded upstream, `validateStatus: () => true` passthrough, SSE streaming supported, graceful 503 when unconfigured
- Clean migration: old `GENAI_URL` sidecar still works with a deprecation warning
- `proxy: false` on the STS call is an intentional, documented corporate-proxy bypass
- Logs carry service/region/auth-mode but never token values
- Proper deploy story this time: `backend/.env.example`, `instance-setup/.env.prod.sample`, and a 124-line `docs/genai-service.md`

### 🟡 Minor

- No Jest tests for the three new services (`externalForward`, `genai`, `serviceToken`) — the backend has a runner, and `serviceToken` especially deserves tests (the fallback path being exactly where tests would have caught the blocker)
- `upstreamPath` is client-controlled and appended to the base URL — confined to the configured host, just noting the contract

**Verdict: request-changes** — one blocker (remove the subscription-key fallback), one decision (RBAC/rate-limit). ECA check failing (standing exception).

7inh:
> ## 🔴 Blocker: Subscription Key Fallback — Acknowledge + Documented Decision

-> removed fallback
---

## 🟠 RBAC/Rate Limit on `/v2/genai/token`

No additional permission gate is needed here. The GenAI service in this deployment is hosted and operated by the customer themselves — it is not a shared billable resource that requires per-user access control. The customer's explicit requirement is that voice input is available to all signed-in users. Adding a `USE_GEN_AI` permission check would conflict directly with that requirement.

If a central platform ever hosts a shared-cost GenAI service, the right place to add that gate is consistently across all `/v2/genai/*` routes (including `/generation`, which is the expensive call and equally ungated today). A review that gates `/token` but not `/generation` would create an inconsistent security posture. We are happy to add uniform permission gating in a follow-up when `/v2/genai/*` cost attribution becomes a concern.

---

## 🟡 `upstreamPath` — SSRF Surface Explanation

The reviewer correctly notes that `upstreamPath` is client-controlled and appended to the base URL. Here is the full containment:

1. `resolveGenAIPath` in `genai.service.js` (`backend/src/services/genai.service.js`) handles only three route shapes: `/generation[/env]`, `/profiles[/id[/env]]`, and a literal passthrough for everything else (e.g. `/actuator/info`, `/conversations/…`). The query string is stripped before the path is resolved.
2. The base URL is `EXTERNAL_GENAI_URL`, which is a fixed operator-configured value baked into the config at startup — the client cannot influence it.
3. The combined URL is therefore always `<fixed-operator-host><sanitized-path>`. A client cannot redirect the request to a different host; the worst they can do is request an arbitrary path on the already-trusted GenAI service — which is a valid and intended use.

This is the same contract as a standard reverse proxy with a fixed upstream. No SSRF is possible because the host is not client-controlled.

NhanLuongBGSV:
## Verification of the fixes — APPROVE

**🔴 Blocker (subscription-key fallback) — fixed and verified.** The fallback is gone from `serviceToken.service.js`; the new contract is stated in the code: *"The subscription key stays on the server; clients receive only the STS token + region."* Failure semantics are now correct: **503** when unconfigured, **502** when Azure STS itself fails — no path returns the key. And the added unit tests (`backend/tests/unit/services/serviceToken.service.test.js`) cover exactly the previously-broken territory: unknown service → 400, missing key/region → 503, STS success shape, and STS failure rejections. Tests where they'd have caught the original bug — the right response to review.

**🟠 RBAC/rate-limit — accepted as a documented deployment decision.** The reasoning holds for the deployment model: customer-hosted GenAI per instance (the Azure key is the deployment's own cost), voice-for-all-signed-in-users is the stated requirement, and gating `/token` while `/generation` stays open would be inconsistent. Noted for the record: if a shared, centrally-billed GenAI deployment ever exists, uniform gating across `/v2/genai/*` is the agreed follow-up.

**🟡 upstreamPath — containment confirmed.** Fixed operator host + the three sanitized path shapes = standard fixed-upstream reverse-proxy contract; no host control reaches the client. No SSRF.

**Verdict: approve** — the secret-exposure path is eliminated with correct error semantics and tests, and both design questions are settled with documented rationale. (ECA check remains the standing policy exception.)

NhanLuongBGSV:
**Follow-up: independent adversarial verification — CONFIRMED**

A second-pass review tried to refute the approval above and could not:

- **Fallback removal verified in the interdiff** — `subscriptionResult` / `auth: 'subscription'` deleted tree-wide; the Azure key is used in exactly one place (the `Ocp-Apim-Subscription-Key` header to the STS endpoint), which the shipped test asserts verbatim
- **Tests executed:** `serviceToken.service.test.js` → **5/5 pass**. Full unit sweep at head: 61 passed / 5 failed — all five failures pre-existing and untouched by this PR (stale password-validator expectations ×3; a plugin test requiring live Mongo ×2)
- **Commit `5e2ce50f` audited** despite its name: it generates nothing persistent (it was actually the commit that introduced the original fallback, subsequently removed by `e061e953`; minting is stateless — no DB model, no stored secrets)
- **Auth:** `router.use(auth())` first middleware with `optional=false` — everything under `/v2/genai/` is mandatory-auth; the no-RBAC stance matches the accepted customer-hosted rationale
- **Secret sweep:** the device token leaves the process only as the upstream `Authorization` header — never logged, never in responses; upstream response headers are dropped (no echo path)

One operational nit for the record: `.env.example` / `.env.prod.sample` ship `EXTERNAL_GENAI_*` **uncommented with placeholders** — an unedited copy makes the service consider itself configured (upstream failures instead of a clean 503). Worth commenting those out in a follow-up.

Merge-ready from review side. (ECA remains the standing policy exception.)

PR #647, opened 2026-08-18
Feat/Add hook to async model vss to external service

7inh:
# Summary

Add optional, pluggable external data sync for Vehicle Model / Extended API / CVI changes. Local CUD still succeeds when the external endpoint is down; the UI shows a simple warning instead of failing the save.

# Motivation

* **What problem does it solve?**
  Deployments need to push AutoWRX model/API changes to a customer-specific external system, where the URL, method, and payload may differ per client.

* **What was difficult or missing before?**
  There was no shared hook for Model / ExtendedApi / API CUD. Wiring sync into each service/controller would have been hard to review and easy to miss on bulk operations such as replace vehicle API, which can involve hundreds of inserts.

* **Why is this approach useful?**
  A Mongoose plugin plus a replaceable handler file keeps the core business logic small. Clients can swap `backend/sync-handlers/default.handler.js` at deployment time. The JWT from the current user is forwarded on outbound calls. Failures are not retried and do not roll back MongoDB.

# Changes

* Added `backend/src/sync/index.js` with:

  * Global Mongoose hooks
  * `AsyncLocalStorage` request context
  * JWT passthrough
  * `X-Sync-Warning` header
  * Single-attempt outbound calls with `proxy: false`
* Added `backend/sync-handlers/default.handler.js` as the per-deployment handler:

  * `CREATE`
  * `UPDATE`
  * `DELETE`
  * `BULK_REPLACE`
  * Optional `CLIENT_SYNC_BASE_URL`
* Updated `createModel` / `replaceApi` to use `runWithSkipSync` during bulk inserts, followed by a single `BULK_REPLACE` sync. This prevents replace/import operations from spamming the external API.
* Updated CORS to expose `X-Sync-Warning`.
* Added a frontend Axios interceptor that:

  * Shows a user-facing toast for sync warnings on mutating requests.
  * Ignores `GET` requests.
  * De-duplicates toasts.
* Updated model deletion to wait briefly before redirecting so the warning can be read.
* Refactored `createModelService` to return the model ID string so import/create navigation does not become `/model/[object Object]`.
* Merged current `main` changes:

  * Email masking
  * Visibility filters
  * ZIP import flow

# Behavior

## Before

* Model/API CUD only updated MongoDB. No external notification was sent.
* Import/create with many extended APIs had no bulk-sync path.
* There was no user feedback when a downstream system needed the change.

## After

* After a successful local write, the backend calls the deployment handler once.

* Create-with-APIs / replace API uses a single `BULK_REPLACE` sync.

* If the external call fails or the host is unreachable:

  * The MongoDB write remains successful.

  * The UI shows:

  > Your changes were saved, but could not be synced to the external system.

* Opening Vehicle API (`GET`) does not show a leftover sync toast from a previous create/update.

* Most deployments can keep the default handler or use a no-op handler.

* Customers with an external integration can replace the handler file with their deployment-specific implementation.

# Testing

* [ ] Unit tests
* [ ] Integration tests
* [x] Manual testing

**Tested with:**

* Python mock server: `C:\repo\sync-mock-server` on `localhost:5050`
* `CLIENT_SYNC_BASE_URL=http://localhost:5050`
* Create / update / delete model
* Replace Vehicle API:

  * One `POST .../apis/replace`
  * No N-per-API calls
* Sync failure scenarios:

  * `-err all`
  * `-err POST`
  * Server stopped
* Verified sync failures show a toast on CUD requests but not on Vehicle API `GET`.
* Verified model import/create success message and list after reload.

# Breaking Changes

None.

# Additional Notes

* The handler is always loaded from `backend/sync-handlers/default.handler.js`. Replace that file per client; do not add `DATA_SYNC_*` environment flags.
* Sync is a single attempt with no queue or retry.
* Corporate `HTTP_PROXY` is bypassed for the outbound Axios client.
* Warnings are attached only to mutating HTTP methods so a slow or failed create sync cannot leak into a later `GET` such as the Vehicle API tab.
* Create/replace waits for the single sync attempt before responding, with a 30-second timeout if the endpoint is unavailable. This is intentional so the toast belongs to the same request.
* **Follow-ups:**

  * Consider optional fire-and-forget sync if 30-second create latency becomes unacceptable.
  * Add automated tests around the Mongoose plugin and `X-Sync-Warning` header.


NhanLuongBGSV:
## Review: pluggable external data sync hook (Model / ExtendedApi / CVI)

**What it is:** a global Mongoose plugin on the Model/ExtendedApi/Api schemas that fires a single outbound sync call on CUD, with per-request context via `AsyncLocalStorage`, an `X-Sync-Warning` response header on failure (surfaced as a UI toast), and a per-deployment handler file (`sync-handlers/default.handler.js`, no-op by default, with an example to copy).

### ✅ Solid design choices

- **Default handler is a no-op** — merged, nothing changes until a deployment deliberately wires a handler; the example file + docs make the extension point obvious
- **Failure isolation is right:** local writes always succeed; failure produces a header + toast, not a failed save; no hidden retries
- CORS `exposedHeaders` handled so the browser can actually read the warning header; frontend interceptor + toast util are clean
- Schema-plugin approach avoids wiring sync into every controller (the stated motivation) — and `runWithSkipSync` exists for internal/system saves

### 🟠 Two design decisions to confirm before merge

1. **The user's platform JWT is forwarded verbatim to the external endpoint** (`Authorization` passthrough via the request context, documented in the handler header). That sends a credential *valid against AutoWRX itself* off-box to a customer-controlled system. It's disclosed and intentional — but please make it an explicit opt-in (env flag, default off) or mint a dedicated scoped sync token instead. If passthrough stays default-on, say so in the deployment docs prominently.
2. **Sync is awaited in the request path** (needed so the warning header can ride the response) with `SYNC_TIMEOUT_MS = 30000`. With any real handler, a slow external endpoint couples the user's save latency to it — worst case a 30-second hang on model/API saves. Since the no-op default makes this invisible today, please shorten the default timeout materially (2–3 s) and/or document the latency contract for handler authors.

### 🟡 Minor

- Single attempt, no retry/queue: a failed sync means the external system silently misses that change (toast only). Disclosed in docs — fine for v1, worth a known-limitations note
- `proxy: false` bypasses corporate HTTP proxies for the outbound call — intentional and documented, but deployments that route egress through a proxy will silently bypass it
- The `model.controller.js` refactor (+105/−52) around the sync insertions deserves a careful self-pass — and the backend has a Jest runner: the sync engine (ALS context, `skipSync`, warning-header emission, handler contract) is very testable and currently untested

**Verdict: request-changes (soft)** — no correctness bug found; the two asks are an opt-in (or replacement) for JWT passthrough and a shorter default timeout. ECA check failing (standing exception).

7inh:
## Awaited Sync (Request Path)

Kept intentionally for v1 so `X-Sync-Warning` can be set on the same mutating response before it is sent (needed for the frontend toast). Documented in `backend/src/sync/index.js` and the handler JSDoc: save/create/delete may wait on one outbound call (timeout `SYNC_TIMEOUT_MS`). Making the timeout env-configurable is a reasonable follow-up if latency becomes an issue.

## Single Attempt / No Retry

Accepted for v1 and documented as a known limitation: one outbound attempt, no queue/retry; local DB write still succeeds. A failed sync can leave the external system behind until a later successful write. The frontend toast is the user-facing signal.

## `proxy: false`

Removed the corporate-proxy bypass. Outbound sync now respects `HTTP_PROXY` / `HTTPS_PROXY` when set.

## Tests

Added `backend/tests/unit/sync/sync.test.js` covering:

* `EXTERNAL_SYNC_DEVICE_TOKEN` on the outbound Axios client
* Proxy not forced off
* `X-Sync-Warning` on mutating failure
* `runWithSkipSync` suppressing outbound sync
* No warning header on `GET`

JWT passthrough was already replaced earlier with `EXTERNAL_SYNC_DEVICE_TOKEN` (dedicated env token).

Happy to iterate on timeout configurability or a retry/outbox if product needs durability beyond best-effort sync.

NhanLuongBGSV:
## Verified — APPROVE

All asks from review are resolved, each verified in the current diff:

- **✅ JWT passthrough → replaced with a dedicated `EXTERNAL_SYNC_DEVICE_TOKEN`** (optional env). The user's platform JWT no longer leaves the box; outbound auth is a deployment-scoped static token, with helpful log guidance for 401/403 causes. This was the main security ask — resolved the right way.
- **✅ Corporate-proxy bypass removed** — `proxy: false` is gone; outbound sync now respects `HTTP_PROXY`/`HTTPS_PROXY`.
- **✅ Awaited-sync latency contract documented in code** — `backend/src/sync/index.js` and the handler JSDoc now state that save/create/delete may wait on one outbound call (`SYNC_TIMEOUT_MS`). Keeping the 30 s constant for v1 is an accepted, documented trade-off (default handler is a no-op, so stock deployments are unaffected); env-configurability noted as the agreed follow-up if it bites.
- **✅ Tests added** — `backend/tests/unit/sync/sync.test.js` covers exactly the right behaviors: device-token attachment (and that `proxy` is *not* forced off), `X-Sync-Warning` on mutating failure, `runWithSkipSync` suppression, and no warning header on `GET`.

Single-attempt/no-retry remains a documented v1 limitation (toast as the user-facing signal) — reasonable for best-effort sync; an outbox/retry is the natural future upgrade if durability requirements appear.

**Verdict: approve.** (ECA check remains the standing policy exception.)

NhanLuongBGSV:
## ⚠️ CORRECTION — supersedes the approval above: request-changes

An adversarial verification pass (real module + real Express, offline) found defects that break the PR's central claim — *"local writes never fail due to sync."* The approval above is withdrawn until F1/F2 are fixed.

### F1 — HIGH: `X-Sync-Warning` header injection turns a successful write into a broken response

`sync/index.js:441` sets the header to `JSON.stringify(warnings)`, where warnings embed the **user-controlled model name** (`formatResourceLabel`) and the **external system's error message verbatim**. Node rejects any header value outside latin1. Reproduced end-to-end: a model named `テスト車両モデル` (or `Modèle — sport`, or any emoji — `model.validation.js` imposes no charset restriction) plus one sync failure → `ERR_INVALID_CHAR` thrown inside the patched `res.json` **after MongoDB committed** → unusable response → user retries → duplicate model. Fix: sanitize/percent-encode the header value and cap its length.

### F2 — MEDIUM-HIGH: a broken handler file 500s create/replace *after* the write

`loadHandler()` runs at `sync/index.js:257`, **outside** the try at `:268`. Replacing `sync-handlers/default.handler.js` — the feature's advertised customization point — with a file that throws at module load makes `triggerSync` reject (reproduced). The Mongoose hook path is guarded, but the three manual call sites (`model.controller.js:161, 171, 433`) `await triggerSync` bare → every create/replace fails with data already persisted. Fix: move `loadHandler()` inside the try and/or guard the controller sites.

### F3 — MEDIUM: documented onboarding fails to boot

`backend/.env.example` ships `EXTERNAL_SYNC_DEVICE_TOKEN=` **empty**; Joi's `Joi.string()` rejects empty. Literally following `README.md:31` (`cp .env.example .env`) → `Config validation error … not allowed to be empty`. One-line fix: comment it out or use a placeholder.

### F4 — MEDIUM-LOW: empty-array truthiness misroutes sync events

`model.controller.js:156` — `Boolean(extended_apis || custom_apis)`: `[]` is truthy. Importing a ZIP with no extended APIs emits a 0-API `BULK_REPLACE` and **never emits `Model CREATE`**.

### F5/F6 — LOW

~16 new lint errors in the PR's own files; and handler caching contradicts the docs (production caches forever — changing the handler silently requires a process restart / image rebuild; dev re-reads every sync). Also the example handler can exceed the documented 30s (timeout is per-request, not per-sync). PR description still advertises JWT passthrough / `proxy:false` — stale vs the shipped code.

### What did hold (verified)

JWT sweep clean (only `Bearer ${EXTERNAL_SYNC_DEVICE_TOKEN}` leaves the box; never logged), ALS context integrity (no cross-request bleed), plugin blast radius (exactly 3 schemas; no write path bypasses the hooks; only the model-delete `deleteMany` cascade is unsynced, by design), frontend `tsc` clean, no toast loop, and the new tests pass verbatim — though they mock axios/handler entirely, which is precisely how F1/F2 escaped: **the header-serialization and loader paths have zero coverage**.

**Verdict: request-changes** — F1 sanitizer, F2 loader guard, F3 env line, F4 truthiness; a test exercising the real header path (non-latin1 model name + failing handler) would lock F1 shut.

7inh:
### Key Changes & Fixes

**Addressed Request-Changes Findings**

* **F1:** `X-Sync-Warning` is now percent-encoded and length-capped; `setHeader` logic is wrapped in a `try/catch` block. Added a regression test using a non-Latin1 model name to assert that Node does not throw `ERR_INVALID_CHAR`.
* **F2:** Moved `loadHandler()` inside `triggerSync`'s `try` block. A broken custom handler now yields a soft sync failure/warning instead of rejecting `create`/`replace` operations after the MongoDB write succeeds.
* **F3:** Commented out the empty token entry in `.env.example`; updated Joi validation to allow empty strings (`''`).
* **F4:** Empty `extended_apis` or `custom_apis` arrays no longer improperly route to `BULK_REPLACE`.

**Outbound Proxy Behavior Follow-up**

* **Issue:** After switching to honor system HTTP/HTTPS proxies, calls to docker-internal endpoints (e.g., `http://genai:8080/profiles/{id}`) were being routed through `rb-proxy-apac.bosch.com`. Since that hostname is not resolvable within the container, requests failed with `ENOTFOUND` despite having valid URLs and paths.
* **Fix:** Reverted default Sync Axios behavior back to `proxy: false` so internal container targets bypass the corporate proxy by default. Added an opt-in flag (`EXTERNAL_SYNC_USE_PROXY=true`) for deployments where the sync URL target is truly external and requires corporate egress.
* **Additional Improvements:**
* Enhanced `ENOTFOUND` error context in application logs.
* Documented the new flag in the handler and `.env.example`.
* Explicitly set `EXTERNAL_SYNC_USE_PROXY=false` in the deployment environment.



> **Note:** The handler contract remains unchanged—it still sends `PUT`/`DELETE` requests to `/profiles/{modelId}` with the computed VSS payload.

Issue #651, opened 2026-08-19
[monogodb obsolete] Staging MongoDB crash-loop: WiredTiger index corruption trips TTLMonitor every 60s


NhanLuongBGSV:
## Summary

A staging-environment MongoDB instance has been **crash-looping roughly once per minute for several days** (~7,000 restarts and counting). The application kept working between crashes, so it went unnoticed by ping-based health monitoring. Root cause is diagnosed and both a quick mitigation and a proper repair are defined.

*(Cross-ref: operational tracking in #649)*

## Root cause

On-disk **WiredTiger corruption in a single index file**:

1. MongoDB's **TTLMonitor** background thread runs a pass **every 60 seconds**
2. Each pass tries to open a cursor on the corrupt index file → `WT_PANIC: WiredTiger library panic` → fatal assertion → process aborts
3. The container restart policy revives it; startup reports an **unclean shutdown** and recovers from the last clean checkpoint
4. Sixty seconds later the next TTL pass hits the same corrupt file → crash again

The one-per-minute cadence is the fingerprint: restart count ÷ elapsed days ≈ exactly one per TTL pass. A catalog scan found **no TTL indexes defined at all**, so the corrupt file is very likely an orphan from a long-dropped index — nothing of value depends on it.

## Why it went unnoticed

The health check pings the database; the ping succeeds for most of each minute between crashes, so the instance reported healthy ~95% of the time. Restart loops are invisible to liveness probes.

## Impact

- Application functionality survives (driver reconnects each cycle), but:
- **Every crash is an unclean shutdown** — journal replay from the last checkpoint, so writes near the end of each cycle can be lost; thousands of such events accumulate consistency risk
- Connected clients take a brief disconnect every minute
- **Restarting cannot fix on-disk corruption** — the loop is indefinite until repaired

### Quick mitigation (seconds, reversible, zero data risk)

Disable the TTL monitor — it is the only code path that opens the corrupt file, and with no TTL indexes defined, nothing needs it:

```yaml
# db service command:
command: ["--setParameter", "ttlMonitorEnabled=false"]
```

Stops the crash loop immediately; remove the line to revert. Suitable as an interim measure while scheduling the proper repair.

### Proper repair (~15–20 min window)

1. Take a fresh dump during an up-window; **restore-test it in a throwaway instance** — prove the backup is good before touching anything
2. Stop the app, remove the DB container and its **volume** (corruption lives on disk)
[...]

PR #652, opened 2026-08-25
chore(instance-setup): move template off EOL mongo 4.4 to 7.0

NhanLuongBGSV:
## What

- `instance-setup/docker-compose.prod.yml`: `mongo:4.4.6-bionic` → `mongo:7.0`
- Healthcheck switched from the legacy `mongo` shell to `mongosh`
- Commented tools bind-mount + instructions for `mongodump`/`mongorestore` (not bundled in 6.0+ images)
- In-file note: existing 4.4 volumes migrate via dump → wipe → restore (restoring rebuilds indexes; no stepwise FCV upgrades needed)

## Why

MongoDB 4.4 has been end-of-life since February 2024 — new deployments should not inherit an unpatched engine. Two operational differences with 7.0 are handled inline so the template works out of the box:

1. The legacy `mongo` shell doesn't exist in 6.0+ images — left unchanged, the healthcheck fails, the container reports unhealthy, and the app's `depends_on: service_healthy` gate blocks startup entirely
2. Database tools are no longer in the image — deployments that exec `mongodump` in-container need the documented one-time tools mount

Compatibility: the backend runs Mongoose 8 (Node driver 6), which supports 7.0 — no application changes required.

## How verified

- Template validated on a live staging instance on 2026-08-25: fresh 7.0.30 volume created via dump → restore (95k+ documents, 0 failures, indexes rebuilt), `mongosh` healthcheck green, in-container backup tooling working through the exact tools mount documented here
- YAML validated (`docker compose config`)
- No application code touched; no test surface affected

No docs-capability changes (deployment template only).

### DIFF
--- a/instance-setup/docker-compose.prod.yml
+++ b/instance-setup/docker-compose.prod.yml
@@ -51,16 +51,28 @@ services:
   autowrx-db:
     # platform: linux/amd64  # Commented out for local testing on non-AMD64 machines
     container_name: ${NAME:-prod}-autowrx-db
-    image: mongo:4.4.6-bionic
+    # 4.4 is end-of-life (Feb 2024, no security patches). 7.0 is a drop-in for
+    # this stack (Mongoose 8 / Node driver 6). Upgrading an EXISTING 4.4 volume
+    # in place is not supported — dump from 4.4, wipe the volume, restore into 7.0
+    # (restoring rebuilds all indexes and needs no stepwise FCV upgrades).
+    image: mongo:7.0
     # ports:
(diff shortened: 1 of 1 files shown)


NhanLuongBGSV:
Related issue: #651 — the staging WiredTiger corruption incident that ran on an end-of-life 4.4 instance, and the live dump → fresh-7.0-volume → restore repair (95k+ documents, 0 failures) that served as the validation for this template change. The migration path documented in the template file here is exactly the sequence executed in that incident.

PR #657, opened 2026-08-26
fix (ux): prototype tab code typing behavior

7inh:
# Summary

Preserve the Monaco editor cursor position when prototype code auto-saves. Previously, saving raced with in-progress typing and forced the editor value to an older snapshot, which jumped the caret to the end of the file.

# Motivation

While editing code in the Prototype Code tab, the cursor jumped to the end of the file whenever an auto-save completed. That made continuous editing frustrating and error-prone (new keystrokes landed at the wrong place).

`saveCodeToDb` always called `setCode` after a successful API write, and the `[prototype]` effect always synced `prototype.code` back into editor state—including when that update was just our own save echoed through `setActivePrototype`. If the user typed during the async save, Monaco received a stale controlled `value`, ran `executeEdits` with `forceMoveMarkers`, and moved the cursor to the end.

## Why is this approach useful?

It keeps local editor state authoritative during save:

- After a successful save, only `savedCode` / refs / the store are updated—not the live `code` state.
- The prototype effect applies `setCode` only on initial load or true external (plugin/AI) updates.
- A `codeRef` avoids stale closures so auto-save persists the latest editor content.

Cursor position and mid-save edits are preserved without changing the save cadence or UX.

# Changes

- Updated `PrototypeTabCode.tsx` so successful saves no longer call `setCode`, avoiding overwrite of newer in-editor content.
- Updated the `[prototype]` effect to distinguish initial load, external updates, and own-save echoes; own-save echoes no longer clobber editor state.
- Added `codeRef` to track the latest editor value for auto-save and race-safe comparisons.
- Refactored save/sync flow so dirty detection still works when the user types during an in-flight save (next tick persists the newer content).

# Behavior

## Before:

- Edit in the middle of the file and keep typing.
- Auto-save (~3s) or blur-triggered save completes.
- Cursor jumps to the end of the file; characters typed during the race can appear at the wrong location.

## After:

- Edit in the middle of the file and keep typing through auto-save.
- Save completes in the background.
- Cursor stays where the user is typing; any characters entered during the save remain in place.
- Unsaved edits typed during the save are still detected as dirty and persisted on the next save cycle.

# Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Tested with:

- Manual editing in the Prototype Code tab while auto-save ran.
- Continuous typing across multiple auto-save intervals to reproduce the prior race.
- Confirmed cursor no longer jumps to end of file after save.

# Breaking Changes

None.

# Additional Notes

- External code updates (plugin/AI) still flow through the prototype effect and replace editor content intentionally; that path is unchanged in purpose.
- Store `prototype.code` may briefly lag local editor content if the user types during an in-flight save; the next auto-save reconciles this.
- `CodeEditor.tsx` was left as a controlled Monaco editor; the fix is entirely in the save/sync ownership logic in `PrototypeTabCode.tsx`.
- Possible follow-up: apply the same “don’t clobber local state on own save” pattern to `ProjectEditor` if it shows a similar cursor/content reset.

### DIFF
--- a/frontend/src/components/organisms/PrototypeTabCode.tsx
+++ b/frontend/src/components/organisms/PrototypeTabCode.tsx
@@ -106,7 +106,6 @@ const PrototypeTabCode: FC = () => {
     ],
     shallow,
   )
-  const [savedCode, setSavedCode] = useState<string | undefined>(undefined)
   const [code, setCode] = useState<string | undefined>(undefined)
   const [ticker, setTicker] = useState(0)
   const [activeTab, setActiveTab] = useState('api')
@@ -177,9 +176,7 @@ const PrototypeTabCode: FC = () => {
 
   useEffect(() => {
     if (!prototype) {
-      setSavedCode(undefined)
       savedCodeRef.current = undefined
-      codeRef.current = undefined
       setCode(undefined)
       setEditorType('code')
       setPreviousCode(undefined)
@@ -200,8 +197,6 @@ const PrototypeTabCode: FC = () => {
         setShowDiff(getDiffVisibleFromSession(prototype.id))
       }
       setCode(prototypeCode)
-      codeRef.current = prototypeCode
-      setSavedCode(prototypeCode)
       savedCodeRef.current = prototypeCode
     } else if (isExternal) {
       // External (plugin/AI) update — capture previous code for diff
@@ -210,13 +205,9 @@ const PrototypeTabCode: FC = () => {
       setPreviousCode(savedCodeRef.current)
       setShowDiff(true)
       setCode(prototypeCode)
-      codeRef.current = prototypeCode
-      setSavedCode(prototypeCode)
       savedCodeRef.current = prototypeCode
-    } else {
-      // Own save echoed via setActivePrototype — do not clobber in-progress edits
-      setSavedCode(prototypeCode)
     }
+    // Own save echoed via setActivePrototype — do not clobber in-progress edits
 
     const newEditorType = getEditorType(
       isInitial || isExternal ? prototypeCode : (localCode ?? prototypeCode),
@@ -230,24 +221,28 @@ const PrototypeTabCode: FC = () => {
       // Explicit save from editor (Save All or Ctrl+S) — always persist, do not skip
       dataToSave = codeToSave
     } else {
-      // Periodic auto-save — skip if unchanged
+      // Periodic / blur auto-save — skip if unchanged (use ref to avoid stale skip)
       const latest = codeRef.current
-      if (latest === undefined || latest === savedCode) return
+      if (latest === undefined || latest === savedCodeRef.current) return
       dataToSave = latest
     }
 
-    if (!prototype?.id) return
+    const prototypeId = prototype?.id
+    if (!prototypeId) return
 
     try {
-      await updatePrototypeService(prototype.id, {
+      await updatePrototypeService(prototypeId, {
         code: dataToSave || '',
       })
       // Mark as saved without setCode — preserves cursor and any edits typed during await
-      setSavedCode(dataToSave)
       savedCodeRef.current = dataToSave
-      const newPrototype = JSON.parse(JSON.stringify(prototype))
-      newPrototype.code = dataToSave || ''
-      setActivePrototype(newPrototype)
+      // Merge only `code` into the *current* store prototype. Cloning the render-closure
+      // prototype would revert fields (widget_config, extend, …) updated elsewhere
+      // while the editor was mounted (e.g. dashboard template apply → blur-save).
+      const current = useModelStore.getState().prototype
+      if (current?.id === prototypeId) {
+        setActivePrototype({ ...current, code: dataToSave || '' })
+      }
     } catch (err) {
       console.error('Error saving code:', err)
     }
@@ -426,7 +421,7 @@ const PrototypeTabCode: FC = () => {
               prototypeName={prototype.name}
               onChange={(data: string) => {
                 setCode(data)
-                // Do not set savedCode here — only when we actually persist (saveCodeToDb)
+                // Do not advance savedCodeRef here — only when we actually persist (saveCodeToDb)
                 // so that Save All / Ctrl+S still trigger the API when there are unsaved changes
               }}
               onSave={async (data: string) => {
--- a/frontend/src/components/organisms/PrototypeTabCode.tsx
+++ b/frontend/src/components/organisms/PrototypeTabCode.tsx
@@ -361,11 +361,12 @@ const PrototypeTabCode: FC = () => {
                           setShowDiff(true)
                         }
                         setCode(newCode)
-                        // Immediately advance savedCodeRef so that the next generation
-                        // (whether via this dialog or via plugin) correctly diffs against
-                        // this result rather than the stale pre-AI code.
-                        savedCodeRef.current = newCode
                         setIsOpenGenAI(false)
+                        // Persist immediately. Do not advance savedCodeRef here — that would
+                        // mark the editor "clean" and skip auto/blur save, leaving AI code
+                        // unsaved and vulnerable to being clobbered by a store refetch.
+                        // saveCodeToDb advances savedCodeRef only after the API succeeds.
+                        void saveCodeToDb(newCode)
                       }}
                     />
                   </Suspense>
--- a/frontend/src/components/organisms/PrototypeTabCode.tsx
+++ b/frontend/src/components/organisms/PrototypeTabCode.tsx
@@ -127,6 +127,8 @@ const PrototypeTabCode: FC = () => {
   const [showDiff, setShowDiff] = useState(false)
   // Ref to track last code saved by us — used to detect external (plugin/AI) code changes
   const savedCodeRef = useRef<string | undefined>(undefined)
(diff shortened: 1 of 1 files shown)


tri2510:
## Review — verified against the code, state machine walked

**The bug is real and the diagnosis is correct.** Pre-fix, two paths force-set the editor value after a save: `saveCodeToDb` did `setCode(dataToSave)` on API success, and the `[prototype]` effect unconditionally did `setCode(prototypeCode)` when the store echoed the save back. With typing during the `await`, both write an *older* snapshot over newer keystrokes — clobbering edits and resetting the caret. The removed lines confirm both.

**The fix addresses the root cause, not the symptom:**
- `codeRef` mirrors the latest editor value — `saveCodeToDb` reads it instead of the closure-captured `code`, so keystrokes during the save `await` are what gets saved next tick, not a stale snapshot
- Save success updates only the saved-markers (`setSavedCode`/`savedCodeRef`) — **no `setCode`**, so the editor value prop never changes and Monaco keeps its caret
- The `[prototype]` effect is now a proper three-way: initial (session diff restore + set), external change (diff capture + set — pre-existing semantics preserved), and **own-save echo → saved-markers only** ✓
- `getEditorType` uses the local code in the echo case — correct, the local value is the newer truth

**Correctness notes from the walkthrough:**
- The key race — typing during the save await — now resolves correctly: `codeRef` advances, save completes with the older-but-valid payload, echo hits the own-save branch, editor untouched, next auto-save persists the typed-ahead delta
- External-change handling unchanged (external still wins + shows the diff banner — same as before, and desirable)
- `savedCodeRef.current!` assertion is safe (only reachable when non-initial)
- The `codeRef` reset on prototype-clear matches `savedCodeRef`'s lifecycle — no cross-prototype leakage

**Verification honesty:** no frontend test runner exists in-repo (known gap), so this is a code-verified review plus a state-machine walkthrough — not a browser-executed one. The change is UX-scoped, single component, no security/auth surface touched, and preserves all existing semantics — proportionate risk. Recommend the author's manual QA note (typing continuously across ≥2 auto-save cycles) land in the PR body, and it'll be user-visible on the next staging deploy for real confirmation.

**Verdict: approve** (ECA remaining the standing exception, as with the batch).

tri2510:
## Independent verification — approval CONFIRMED, plus one finding to consider fixing in this PR

A second-pass review (state-machine simulation of the old vs. new logic — 15 scripted interleavings + 8,000 fuzz streams, plus direct inspection of the `@monaco-editor/react` 4.7.0 source) confirms the fix, and finds it **better than claimed**:

- **The old bug was worse than a caret jump**: typing during an in-flight save was *permanently lost* — the echo reverted the editor, then `code === savedCode` meant it never re-saved. Fuzz invariant "typing never reverted": old logic violated **1,477 / 4,000** runs; new logic **0 / 4,000**, with zero post-idle convergence failures
- **Caret mechanism confirmed** in the monaco wrapper source: a changed `value` prop triggers a full-range `executeEdits(..., forceMoveMarkers: true)` — exactly what the removed `setCode` invoked
- `tsc --noEmit` passes; zero new lint issues (91 pre-existing on both old and new)
- **Two unclaimed bonus fixes**: refetch-while-dirty (window refocus) no longer discards edits, and out-of-order saves converge to the newest content instead of the oldest
- External-change semantics (diff banner, `previousCode` baseline) verified unchanged

### F1 (Low–Moderate) — the PR activates a previously-dead blur-save path that echoes a mount-frozen prototype clone

`onBlur={saveCodeToDb}` (`PrototypeTabCode.tsx:461`) was effectively dead before this change — the frozen closure's `code === savedCode` always tripped the skip guard. The live `codeRef` makes blur-save actually fire (good for persistence), but the success handler (lines 248–250) does `setActivePrototype(JSON.parse(JSON.stringify(prototype)))` with the **mount-time** `prototype`, reverting every other prototype field in the store that changed since the editor mounted.

Failure chain: editor mounts → dashboard applies a template (store + API updated) → user types in the code editor → blurs → blur-save PUT succeeds → the echo clones the mount-time prototype → the store's `widget_config`/`extend` revert → the dashboard shows stale config → if the user then presses the dashboard's Save, the reverted config is **persisted to the DB**.

The echo itself doesn't touch the DB and the store self-heals on refetch, so this isn't a merge blocker — but since this PR is what wakes the path, it's ideally fixed here: **merge only the `code` field into the current store state for the echo** (or keep a `prototypeRef` alongside `codeRef`).

### Also noted (not blockers)

- F2 (theoretical, sub-ms): the initial/external branches write `codeRef.current` directly *and* `setCode`; a keystroke restoring the exact prior content before the effect commits can leave `codeRef` at the branch value. Purist fix: let the `[code]` effect be the sole writer of `codeRef`
- Three pre-existing issues unchanged by this PR, for the record: a direct P1→P2 store switch can misfire the diff banner with P1's code as baseline; AI-generated code is silently discarded if a store refresh lands between generation and the next auto-save tick; a save resolving after navigation pushes a stale full-prototype clone (F1 is the newly-live instance of this older pattern)

**Verdict stands: approve** — with F1 recommended for this PR since the PR activates the path.

7inh:
Fixed in this PR:

After a successful save, we no longer use `JSON.parse(JSON.stringify(prototype))` from the render closure.

Instead, we now:

* Read `useModelStore.getState().prototype`.
* Merge only `{ code }` when the ID still matches.

This prevents dashboard updates made while the editor is open — such as `template apply → widget_config/extend` — from being reverted by a blur/ticker save.

Also applied the F2 suggestion: `codeRef` is now written only from the `[code]` effect.

tri2510:
## Re-review at `9733e073` — F1/F2 fixed exemplarily, but the simplification introduces one regression

### ✅ F1 — fixed better than asked

The save echo now reads `useModelStore.getState().prototype` (current store, not the render closure), guards on `current?.id === prototypeId`, and merges **only `{ code }`**. The dashboard-revert chain is dead — and the ID guard additionally protects against a save resolving after prototype navigation (a pre-existing stale-echo pattern, fixed for free here).

### ✅ F2 — fixed the purist way

`codeRef` now has exactly **one** writer in the file (inside the `[code]` effect) — verified programmatically. The theoretical race is structurally gone.

### 🔴 New finding — AI-generated code is no longer persisted

Deleting the `savedCode` **state** and making `savedCodeRef` the sole comparison removes a distinction that was load-bearing:

- **On main**: the SDV ProtoPilot dialog's handler does `setCode(newCode)` + advances `savedCodeRef` (for diff-baselining) — but the auto-save guard compared the **state**, which the handler never touched → AI code registered as dirty → **auto-save persisted it**
- **At this head**: the guard is `codeRef.current === savedCodeRef.current`, and the handler advances `savedCodeRef` to the same new code → the guard sees "clean" → **auto-save and blur-save both skip**
- The dialog itself never persists (`DaGenAI_Python`'s apply button only invokes the callback — no API write; verified)
- Consequence: AI-generated code applied via the dialog is **never saved** unless the user types something more afterward (dirtying the guard) or explicitly Ctrl+S/Save-All's. Worse, a later store refetch sees the server's older code ≠ `savedCodeRef` → treats it as an *external* change → **clobbers the AI code out of the editor**

**Minimal fix:** in the AI handler, don't advance `savedCodeRef` (the dialog path's diff baseline already comes from the `code` closure in the capture block above it) — or call the save explicitly after applying. Worth checking whether the plugin path relied on that advance before removing it.

**Verdict: request-changes** on this single item — everything else in the PR is correct and an improvement over both main and the first revision.

7inh:
After dropping `savedCode` state, the ProtoPilot handler’s `savedCodeRef.current = newCode` caused the new dirty guard to treat the AI-generated code as already saved. As a result, auto/blur save was skipped.

A later refetch could then be interpreted as an external change and wipe the editor.

### Fix

* No longer advance `savedCodeRef` in the dialog handler.
* Call `saveCodeToDb(newCode)` directly when the user applies the generated code.
* The diff baseline still comes from the pre-apply `code` capture.
* `savedCodeRef` advances only after the API call succeeds.
* Plugin/external updates continue to flow through the store → external branch of the prototype effect.

tri2510:
## Re-review at `be2183e2` — all findings resolved — **approve**

The F3 fix is exactly right, and the inline comment shows the full picture:

- `savedCodeRef.current = newCode` **removed** from the ProtoPilot handler; replaced with `void saveCodeToDb(newCode)` — which takes the **explicit-save path** (the `codeToSave !== undefined` branch always persists, bypassing the skip guard; verified against the function signature)
- `savedCodeRef` writers are now only the four semantically-correct sites (prototype reset / initial / external / **post-API save success**) — nothing marks the editor clean without a persisted write
- The diff-baseline need that originally motivated the advance is handled by the `code` closure in the capture block — unaffected
- Resulting AI flow: `setCode` → `codeRef` sync → explicit persist → `savedCodeRef` advances only after API success → echo merges `{code}` into the current store under the ID guard

**Full review arc on this PR:** caret fix → independent simulation (found F1, confirmed the rest — fuzz 1,477→0/4,000) → F1/F2 fixed better than asked → F3 caught on the over-simplification → fixed correctly. Every round verified from head code; the final state improves on main in every dimension the probes measured (typing preservation, refocus-while-dirty, out-of-order saves, stale-echo protection, AI-code persistence).

**Verdict: approve — merge-ready.** (ECA the standing exception, as with the batch.)

PR #658, opened 2026-08-27
fix(frontend): stop remounting action-button rows while editing label

7inh:
## Summary

Stop deriving each right-nav action button's Draggable `key` from mutable field content (`label`, `hideIcon`, `variant`, `corners`). Use a stable identifier (`btn.builtin ?? btn.plugin ?? index`) instead, so editing the Label field in the Staging/Deploy button config no longer remounts the row on every keystroke.

## Motivation

- What problem does it solve? Typing in the Label input for a staging/deploy action button lost focus and cursor position after every keystroke.
- What was difficult or missing before? The `Draggable`/`Droppable` `key` for each row was `JSON.stringify({ plugin, label, hideIcon, variant, corners })`. Any edit to `label` (or the other fields) changed the key, causing React to unmount and remount the entire `ActionButtonEditor` subtree — including the `<Input>` DOM node — on every character typed.
- Why is this approach useful? React keys should express stable identity for reconciliation, not current content. Keying on position/plugin/builtin keeps identity stable across edits while still being unique enough for drag-and-drop reordering.

## Changes

- `frontend/src/components/organisms/ActionButtonsTab.tsx`: replaced the content-based `key` computation in the row `.map()` with `key-${btn.builtin ?? btn.plugin ?? i}`.
- Removed the identical unused content-based `key` computation inside `Droppable`'s `renderClone` (it wasn't referenced after being built).

## Behavior

Before:
- Typing in the "Label" field for a staging/deploy button caused the input to lose focus/cursor after each keystroke, making it unusable for anything beyond a single character at a time.

After:
- Typing in the "Label" field updates state normally without remounting the row; focus and cursor position are preserved while editing.

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- Read-through of `ActionButtonsTab.tsx` confirming the key was the only thing tying `label` changes to remounts; `npx tsc --noEmit` shows no new type errors touching the file.
- `npm run lint` could not be run in this environment (pre-existing repo issue: ESLint 9 installed but repo still ships an ESLint 8-style config, not related to this change).

## Breaking Changes

None. Purely a React reconciliation-key fix; no external behavior, API, or data shape changes.

## Additional Notes

Change is currently staged but uncommitted on `main`; needs its own feature branch (`fix/action-buttons-label-remount-key`) cut before opening the PR, per repo convention (never commit directly on `main`).

### DIFF
--- a/frontend/src/components/organisms/ActionButtonsTab.tsx
+++ b/frontend/src/components/organisms/ActionButtonsTab.tsx
@@ -1,3 +1,11 @@
+// Copyright (c) 2025 Eclipse Foundation.
+//
+// This program and the accompanying materials are made available under the
+// terms of the MIT License which is available at
+// https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: MIT
+
 import { Button } from '@/components/atoms/button'
 import { Input } from '@/components/atoms/input'
 import { Label } from '@/components/atoms/label'
--- a/frontend/src/components/organisms/ActionButtonsTab.tsx
+++ b/frontend/src/components/organisms/ActionButtonsTab.tsx
@@ -485,13 +485,6 @@ const ActionButtonsTab = ({
             type="RIGHT_NAV_BUTTONS"
             renderClone={(provided, snapshot, rubric) => {
               const btn = localRightNavPlugins[rubric.source.index]
-              const key = `key-${JSON.stringify({
-                plugin: btn.plugin,
-                label: btn.label,
-                hideIcon: btn.hideIcon,
-                variant: btn.variant,
-                corners: btn.corners,
-              })}`
               return (
                 <div
                   ref={provided.innerRef}
@@ -528,13 +521,7 @@ const ActionButtonsTab = ({
                 className="flex flex-col gap-2"
               >
                 {localRightNavPlugins.map((btn, i) => {
-                  const key = `key-${JSON.stringify({
-                    plugin: btn.plugin,
-                    label: btn.label,
-                    hideIcon: btn.hideIcon,
-                    variant: btn.variant,
-                    corners: btn.corners,
-                  })}`
+                  const key = `key-${btn.builtin ?? btn.plugin ?? i}`
 
                   return (
                     <Draggable key={key} draggableId={key} index={i}>


tri2510:
## Review — fix verified correct; one addition needed to clear the red check

### ✅ The fix is right

- **Root cause correctly diagnosed**: the row key was `JSON.stringify({plugin, label, hideIcon, variant, corners})` — every Label keystroke changed the key → React remounted the `Draggable` row → the input lost focus/cursor. Same class as the editor-caret bug (#657), and the stable-identity fix (`btn.builtin ?? btn.plugin ?? i`) is the correct shape.
- **`renderClone` verified clean at head**: the removed const wasn't referenced in its JSX (the clone renders via `provided` props) — deleting it breaks nothing.
- **Stability + uniqueness hold**: the add-path dedupes by plugin slug (`:453`), builtin ids come from a fixed set, and the index fallback is unique per position. Two theoretical nits, both noted-not-blockers: a builtin id colliding with a plugin slug would duplicate a `draggableId` (react-beautiful-dnd fails loudly, not silently), and rows with neither id would remount on reorder via the index fallback — unlikely to exist in practice.

### 🔴 Blocker: the license-headers check is red on this PR

`ActionButtonsTab.tsx` carries no MIT header **on main** (it starts with imports), so touching it surfaces the pre-existing debt — the same pattern as the `.agents` files before #648. The check fails on exactly this file. **Please add the standard header block at the top in this PR** (7 lines, `SPDX-License-Identifier: MIT`) — that clears the red check and fixes the debt for good.

**Verdict: approve the logic; merge-ready once the header lands.** (ECA the standing exception.)

7inh:
Confirmed.

Since this PR modifies `ActionButtonsTab.tsx`, it falls within the scope of `check-license-headers.sh`, which checks Added/Copied/Modified files against `origin/main`. The file was missing an SPDX header.

Added the standard 7-line Eclipse/MIT header in `07bce2a0`, matching the format used across the rest of `components/organisms/`.

`./scripts/check-license-headers.sh` now reports:

> `license-headers OK: 1 changed source file(s) all carry the MIT SPDX header.`

tri2510:
Verified at `5c20f948` — header present (standard Eclipse/MIT block, first lines confirmed) and the **license-headers check is green**. Only ECA remains red (the standing exception, as with the batch).

**Verdict: approve — merge-ready.**

PR #659, opened 2026-08-27
feat(frontend): render plugin tabs in the Vehicle API page Tab Bar 

7inh:
## Summary

Add a new `MODEL_API_PLUGINS` site config that lets installed plugins appear as first-class, route-driven tabs in the Vehicle API page's top-level Tab Bar (`ModelApiTabs`), alongside "COVESA API" and custom API set tabs — the same mechanism `VSS_PLUGINS` already provides one level down, inside `ViewApiCovesa`'s internal tab row (opened as a modal).

## Motivation

- What problem does it solve? Plugins configured via `VSS_PLUGINS` only ever open in a modal dialog nested inside the COVESA tab's own internal tabs — there was no way to expose a plugin as a peer of "COVESA API" / a custom API set in the page's main Tab Bar.
- What was difficult or missing before? The plugin-install-check + tab-rendering logic (read site config array → filter valid `{label, plugin}` entries → verify each slug is actually installed) lived inline inside `ViewApiCovesa.tsx`, with no reusable form and no route-level entry point for a plugin tab.
- Why is this approach useful? Reusing the existing route-per-tab pattern (`api`, `api/covesa/:api`, `api/:instance_id`) keeps plugin tabs consistent with how every other Tab Bar entry already behaves — a real URL, browser back/forward support, and inline content instead of a modal.

## Changes

- Added `frontend/src/hooks/useConfiguredPlugins.ts`: extracts the "read site-config array → filter valid entries → keep only installed plugin slugs" logic (previously inline in `ViewApiCovesa.tsx`) into a reusable hook parameterized by config key.
- Refactored `ViewApiCovesa.tsx` to call `useConfiguredPlugins('VSS_PLUGINS')` instead of its inline effect; its own internal plugin tabs/modal behavior is unchanged.
- Added a new `MODEL_API_PLUGINS` predefined site config (`frontend/src/pages/SiteConfigManagement.tsx`), `valueType: 'array'`, editable via the existing generic JSON-array config editor — no bespoke UI needed.
- Added route `api/plugin/:plugin_slug` (`frontend/src/configs/routes.tsx`), rendering `PageVehicleApi`, distinct from `api/:instance_id` (custom API sets) to avoid slug/id collisions.
- `ModelApiTabs.tsx`: accepts a new `plugins` prop and renders one tab per configured/installed plugin, active when `plugin_slug` matches; fixed `isCovesaActive` to also check `!plugin_slug` (previously both "COVESA API" and the active plugin tab could render `active=true` simultaneously on the plugin route, since `instance_id` is undefined there too).
- `PageVehicleApi.tsx`: fetches `MODEL_API_PLUGINS` via the new hook, passes it to `ModelApiTabs`, and renders `PluginPageRender` inline in the content area when the active route matches a configured plugin.
- `PluginPageRender.tsx`: removed a redundant, wrongly-targeted `registration.unmount(containerRef.current)` call in the main effect's cleanup. For `mount()`/`unmount()`-style plugins, the internal `Wrapper` component already calls `unmount()` on the correct node as part of its own `useEffect` cleanup; the outer cleanup was calling it a second time against a different (outer, React-owned) container, which could race with React's own unmount of that subtree and throw `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node` intermittently when switching tabs.

## Behavior

Before:
- Plugins could only be surfaced via `VSS_PLUGINS`, appearing as tabs nested inside `ViewApiCovesa`'s own tab row and opening in a modal dialog.
- No way to add a plugin as a top-level, URL-addressable tab next to "COVESA API" / custom API sets.
- Switching into/out of a plugin tab could intermittently throw a `removeChild` DOM error for `mount()`/`unmount()`-style plugins.

After:
- Site admins can add entries to `MODEL_API_PLUGINS` (`{"label": "...", "plugin": "..."}`); each installed plugin shows up as a tab in the page's top Tab Bar.
- Clicking the tab navigates to `/model/:model_id/api/plugin/:plugin_slug`, the tab renders active, and the plugin loads inline in the content area (not a modal).
- Switching between "COVESA API", custom API sets, and plugin tabs works via normal routing; only one tab shows as active at a time.
- `VSS_PLUGINS` / `ViewApiCovesa`'s internal plugin modal continues to work unchanged.
- Tab switching no longer intermittently throws the `removeChild` DOM error.

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- `cd frontend && npx tsc --noEmit` — clean.
- `cd frontend && npm run build` — succeeds (`vite build`, no new warnings).
- `npm run lint` could not be run in this environment (pre-existing ESLint v9 / legacy `.eslintrc.cjs` mismatch, unrelated to this change).
- Manual: added a `MODEL_API_PLUGINS` entry pointing at an installed plugin, confirmed the tab appears, navigates correctly, renders inline, and that tab-active state and repeated tab switching (including the plugin tab) no longer show the `removeChild` error or a stuck "COVESA API" highlight.

## Breaking Changes

None. `MODEL_API_PLUGINS` is a new, additive, opt-in site config (defaults to `[]`); `VSS_PLUGINS` and existing routes/tabs are unaffected.

## Additional Notes

Targets `main` directly (not stacked). The `PluginPageRender.tsx` cleanup fix is a small, targeted bug fix surfaced by this feature's route-driven tab switching, not a broader rework of the plugin script-injection mechanism.

### DIFF
--- a/frontend/src/components/molecules/ModelApiTabs.tsx
+++ b/frontend/src/components/molecules/ModelApiTabs.tsx
@@ -9,10 +9,11 @@
 import { FC, useMemo } from 'react'
 import DaTabItem from '@/components/atoms/DaTabItem'
 import { useParams } from 'react-router-dom'
-import { TbApi, TbPlus } from 'react-icons/tb'
+import { TbApi, TbFileImport, TbPlus } from 'react-icons/tb'
 import { Button } from '@/components/atoms/button'
 import { useQuery } from '@tanstack/react-query'
 import { getCustomApiSetById } from '@/services/customApiSet.service'
+import { ConfiguredPlugin } from '@/hooks/useConfiguredPlugins'
 
 interface ModelApiTabsProps {
   customApiSetIds?: string[]
@@ -20,6 +21,7 @@ interface ModelApiTabsProps {
   isModelOwner?: boolean
   covesaApiCount?: number // Count of COVESA APIs
   enableCustomApiSets?: boolean
+  plugins?: ConfiguredPlugin[]
 }
 
 const ModelApiTabs: FC<ModelApiTabsProps> = ({
@@ -28,8 +30,13 @@ const ModelApiTabs: FC<ModelApiTabsProps> = ({
   isModelOwner = false,
   covesaApiCount = 0,
   enableCustomApiSets = true,
+  plugins = [],
 }) => {
-  const { model_id, instance_id } = useParams<{ model_id: string; instance_id?: string }>()
+  const { model_id, instance_id, plugin_slug } = useParams<{
+    model_id: string
+    instance_id?: string
+    plugin_slug?: string
+  }>()
 
   // Normalize IDs to strings (handle MongoDB ObjectIds that might be objects)
   // Use useMemo to ensure stable reference and prevent unnecessary re-renders
@@ -61,7 +68,8 @@ const ModelApiTabs: FC<ModelApiTabsProps> = ({
   const sets = enableCustomApiSets ? setQueries.data || [] : []
 
   // Determine active tab
-  const isCovesaActive = !instance_id || instance_id === 'covesa'
+  const isCovesaActive =
+    !plugin_slug && (!instance_id || instance_id === 'covesa')
 
   return (
     <>
@@ -108,6 +116,18 @@ const ModelApiTabs: FC<ModelApiTabsProps> = ({
           </DaTabItem>
         )
       })}
+
+      {/* Plugin tabs */}
+      {plugins.map((p) => (
+        <DaTabItem
+          key={`plugin-${p.plugin}`}
+          active={plugin_slug === p.plugin}
+          to={`/model/${model_id}/api/plugin/${p.plugin}`}
+        >
+          <TbFileImport className="w-5 h-5 mr-2" />
+          {p.label}
+        </DaTabItem>
+      ))}
 
       {/* Plus button to add new instance */}
       {enableCustomApiSets && isModelOwner && onAddInstance && (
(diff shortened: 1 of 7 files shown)


tri2510:
## Review — verified end to end — **approve**

### What was verified

- **Proper reuse, not duplication**: `useConfiguredPlugins` extracts exactly the logic `ViewApiCovesa` carried inline (−47 lines there) — the same shape now powers both `VSS_PLUGINS` (inner tabs) and the new `MODEL_API_PLUGINS` (top-level tabs)
- **Hook correctness**: entry validation, slug dedup, `Promise.allSettled` (one missing plugin can't break the set), cancelled-flag on cleanup, only *installed* plugins resolved
- **Install-check works for every viewer**: `getPluginBySlug` → `/system/plugin/slug/:slug`, which is a **public read endpoint** (verified in the route file) — read-only users see the same tabs; a 404 (not installed) filters out via the rejected promise
- **Security gating holds**: `activePlugin` is looked up *within the admin-configured + installed set* — a crafted URL can't render an arbitrary plugin; the route is scoped under the model tree with `SuspenseProvider`
- **A real bug fixed in passing**: the removed `PluginPageRender` cleanup unmounted against the outer React container *and* the `Wrapper` component unmounts its own node — a double-unmount on mismatched nodes. Verified the `Wrapper`'s `useEffect` cleanup owns `unmountFn?.(ref.current)` on the exact node it mounted into; the removal is correct and the new comment documents why
- **Opt-in by default**: `MODEL_API_PLUGINS` ships `[]` — zero behavior change until an admin configures it; good admin-UI description with example; MIT header on the new hook file; license check green

### Nits (non-blocking)

- `[JSON.stringify(pluginsConfig)]` dep is pragmatic and plays cleanly with the site-config cache (#645); a transient 5xx on the slug check filters the tab until the next pass — acceptable
- No tests (frontend runner gap, known)

**Verdict: approve — merge-ready.** (ECA the standing exception, as with the batch.)

PR #660, opened 2026-08-27
feat(instance-setup): add backup and restore utilities

tri2510:
## What

Two deployment utilities for `instance-setup/`, alongside `up.sh`/`down.sh`:

- **`backup.sh`** — after `up`: dumps the Mongo DB (`backups/mongo-<ts>.archive.gz`) + archives upload/plugin data (`backups/data-<ts>.tar.gz`)
  - **Asks for confirmation before running** (summary of instance/db/paths first); `-y` for automation, `--db`/`--data` to scope
  - Detects a working `mongodump` (built-in on ≤5 images, `/opt/dbtools` mount on 7+ — the mount documented in the compose since #652) and prints one-time setup instructions when absent
  - gzip-integrity check + non-trivial-size signal on the dump; partial files removed on failure
  - Never auto-deletes backups (retention is a deliberate act)
- **`restore.sh`** — restores from those files
  - **Requires typing `RESTORE` in capitals** after a destructive-summary (db `--drop`, data overwritten, post-backup writes lost)
  - Stops the app container for the duration, restarts it after (including on failure paths, via trap) — no writes race the restore
  - Refuses corrupt archives (gzip test), prints post-restore per-collection counts as the verification signal
- Guide section added; `instance-setup/backups/` gitignored

## Why

Deployments had no built-in backup/restore path — the staging incident (#651) and its recovery ran on hand-built scripts. These encode the lessons: mongo-7 has no bundled tools (auto-detect + instruct), dumps from crash-looping instances can't be trusted without verification (integrity checks + counts), restores are destructive and must be confirmed, and the app must be stopped during restore.

## How verified

- `bash -n` clean on both scripts
- The dump/restore tooling paths (`mongodump` builtin → `/opt/dbtools` fallback, `--archive --gzip --drop`, post-restore counts) are the exact commands proven live during the 2026-08-25 staging migration and the 2026-08-26 restore — executed verbatim against a real instance that day
- Not executed against a live instance in this PR (no test stack stood up); the confirmation flows and picker logic are shell-only, syntax-checked

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,5 @@ temp/
 .env.stack.secrets
 !.env.stack.example
 
+# instance backup archives (backup.sh output)
+instance-setup/backups/
--- a/instance-setup/backup.sh
+++ b/instance-setup/backup.sh
@@ -8,7 +8,11 @@
 #   ./backup.sh --data     # data (uploads + plugins) only
 #   ./backup.sh -y         # skip the confirmation prompt (automation)
 #
-# Output: ./backups/mongo-<timestamp>.archive.gz and data-<timestamp>.tar.gz
+# Output: ./backups/autowrx-backup-<timestamp>.zip containing:
+#   - INFO.txt            (instance name, database, source paths, created date)
+#   - .env.prod           (the instance configuration — needed for consistent restore)
+#   - mongo.archive.gz    (the database dump)
+#   - data.tar.gz         (uploads + plugins)
 # Retention is manual — old backups are never deleted automatically.
 
 set -euo pipefail
@@ -30,7 +34,10 @@ PLUGIN_PATH=${PLUGIN_PATH:-./data/plugin}
 DB_CONTAINER="${NAME}-autowrx-db"
 TS=$(date +%Y%m%d-%H%M%S)
 BACKUP_DIR="./backups"
+STAGE=$(mktemp -d)
+trap 'rm -rf "$STAGE"' EXIT
 mkdir -p "$BACKUP_DIR"
+ZIP="$BACKUP_DIR/autowrx-backup-$TS.zip"
 
 DO_DB=1; DO_DATA=1; ASSUME_YES=0
 for arg in "$@"; do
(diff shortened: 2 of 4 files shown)


tri2510:
**Hardened per review feedback:** restore now needs two independent protections — (1) explicit unlock: `./restore.sh --enable` creates a 15-minute token, consumed on use, so scripts/cron/accidental runs are refused; (2) typed confirmation of the exact **instance NAME**, binding the restore to the deployment you're aiming at. Live-testing these caught a real bug: absent-but-optional env vars (legal per compose defaults) killed both scripts silently under `set -e` — fixed and execution-verified (refuse/unlock/wrong-name/consume paths all behave).

tri2510:
**Simplified per operator feedback:** the enable-token + instance-name scheme is gone. Restore protection is now exactly **three typed `RESTORE` confirmations** — one per stage (scope → database overwrite → data overwrite), each showing the details that matter at that point. Execution-verified: abort at stage 1/2/3 all work, and typing all three proceeds into the integrity checks (which correctly refused the zero-byte test archive). Testing also surfaced and fixed two more silent `set -e` aborts in the picker and conditional echos.

tri2510:
**Backup format upgraded per operator request:** `backup.sh` now produces a single **timestamped zip** (`autowrx-backup-&lt;ts&gt;.zip`) containing: **INFO.txt** (instance name, database, data paths, created UTC — the what-came-from-where record), **`.env.prod`** (config at backup time, for consistency; never auto-applied on restore), `mongo.archive.gz`, `data.tar.gz`. `restore.sh` takes the zip (`--backup` or latest), **validates INFO.txt against the current instance/database and refuses cross-instance restores**, then runs the three-confirmation flow. Round-trip execution-tested: backup → wipe → restore → content verified; cross-instance refusal verified.

tri2510:
**Fix:** the previous push (`b50e0df6`) had accidentally clobbered `restore.sh` with `backup.sh`'s content — both files were identical at that head and the confirmation logic was gone. Restored at `1759b19f`: the real restore.sh (three typed confirmations, `*-backup-*.zip` picker, INFO.txt origin validation) — remote verified. Full round-trip re-verified with an **isolated instance NAME** (no shared-host collision): backup → wipe → restore → file content confirmed; cross-instance refusal stands from the earlier test. Apologies for the broken intermediate — caught by re-checking the pushed branch, not just the local tree.

PR #661, opened 2026-09-03
fix: validate origin before redirecting in GitHub auth/SSO callbacks

### LINKED ISSUE #635: Open redirect in GitHub SSO flow
Labels: `security`, `severity:high`
- Problem: Redirect origin from request state/query is used without strict allowlist validation.
- Why it matters: Enables phishing-friendly open redirects after auth flows.
- Fix: Validate post-auth redirect targets against a strict trusted-origin allowlist.

aroh3006:
Fixes #635.

`githubCallback`, `githubSsoStart` and `githubSsoCallback` all take an
`origin` from the query string (SSO's `origin` is carried through the
GitHub round trip inside `state`, but it's the same caller-supplied
value) and redirect to it with no validation:

- `githubCallback`: `res.redirect(`${origin || '...'}/auth/github/success`)`
- `githubSsoStart`: embeds `origin` into `state`, which comes back on
  the callback
- `githubSsoCallback`: `res.redirect(origin)` / `res.redirect(`${origin}?error=...`)`
  on every success and error path

Any of these can be used to send a user to an attacker-controlled URL
right after the GitHub auth flow, since none of them check `origin`
against anything before using it.

`origin` is now checked against the same `CORS_ORIGINS` allowlist
already used for cross-origin requests (`config.isAllowedOrigin`,
factored out of the existing CORS check so both use the same list),
and falls back to the configured client base URL when it isn't on the
list.

**Tests.** Added `backend/tests/unit/config/isAllowedOrigin.test.js`,
which confirms an allowed origin passes, an attacker-controlled one is
rejected (including a suffix-match attempt like
`app.example.com.evil.example`), and a missing origin is rejected.
Ran it against the code before this change and it fails
(`config.isAllowedOrigin is not a function`, since the check didn't
exist yet), and passes with the fix. I didn't run the full `auth.test.js`
integration suite since it needs a live MongoDB instance I don't have
set up locally; happy to if someone can confirm a quick way to run it,
or if a maintainer wants to check before merging.

### DIFF
--- a/backend/src/config/config.js
+++ b/backend/src/config/config.js
@@ -103,6 +103,31 @@ if (envVars.AUTH_PROVIDER === 'platform' && !platformHeaders.email) {
   );
 }
 
+// Shared by CORS and by anything that redirects based on a caller-supplied origin
+// (OAuth/SSO return URLs). Same allowlist, same CORS_ORIGINS patterns, so a host
+// that isn't trusted for CORS isn't trusted as a redirect target either.
+const isAllowedOrigin = (origin) => {
+  if (!origin) {
+    return false;
+  }
+
+  const corsOriginsPatterns = envVars.CORS_ORIGINS.split(',')
+    .map((pattern) => pattern.trim())
+    .filter(Boolean);
+
+  const allowedOrigins = [];
+  corsOriginsPatterns.forEach((pattern) => {
+    try {
+      allowedOrigins.push(new RegExp(`^http://${pattern}$`));
+      allowedOrigins.push(new RegExp(`^https://${pattern}$`));
+    } catch (e) {
+      console.error(`Invalid CORS origin pattern: ${pattern}`, e);
+    }
+  });
+
+  return allowedOrigins.some((pattern) => pattern.test(origin));
+};
+
 const config = {
   env: envVars.NODE_ENV,
   port: envVars.PORT,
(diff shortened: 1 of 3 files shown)


tri2510:
@aroh3006 — reviewed; this is a clean, correct open-redirect fix. **Approve.**

### Verified

- **The vulnerability is real** (fixes #635): `githubCallback`, `githubSsoStart`, and `githubSsoCallback` all redirected to caller-supplied `origin` (carried through the GitHub round-trip inside `state`) with no validation — a textbook open redirect on the auth path
- **The fix reuses the right allowlist**: `isAllowedOrigin` is factored out of the existing CORS middleware (same `CORS_ORIGINS` patterns, same http/https regex construction — I verified the middleware now calls the shared function, so CORS and redirect trust can't drift apart)
- **Fallback is safe**: disallowed origins fall back to `config.client.baseUrl` rather than erroring — no broken flows for legitimate misconfigurations, no user-visible change for allowed origins
- **Tests run and pass** (3/3 on the branch): allowed origin ✓, evil suffix `app.example.com.evil.example` ✓ (the classic regex-suffix bypass — correctly rejected because the anchored `^https://pattern$` construction), undefined/empty ✓
- Bonus: the refactor leaves the CORS middleware semantically identical — behavior change is scoped to the three redirect sites only

### One ask, non-blocking

ECA is failing — sign at https://www.eclipse.org/legal/eca/ with the email your commits use and this can merge.

PR #662, opened 2026-09-07
fix(frontend): align create model/prototype dialogs and select UX

7inh:
## Summary

Improves the Create New Model and New Prototype dialogs so they share consistent header/field styling, autofocus the name inputs, reuse one create-model dialog component, and fix select dropdown UX (width, outside-click/Escape dismiss, focus-ring clipping).

Create Model and New Prototype flows looked and behaved inconsistently: different title styles, mixed label/input components, missing dialog titles in some entry points, selects that overflowed the dialog or would not close on outside click/Escape, and focus rings that were cropped by overflow containers.

- Create New Model was copy-pasted across Home, Model List, Home buttons, and `/new-prototype` with slightly different dialog props.
- New Prototype rendered its own in-form title instead of the shared dialog header.
- Nested Radix Selects inside `DaDialog` did not reliably dismiss on outside click or Escape.
- Long model names widened the select menu past the field/dialog.
- Focus `box-shadow` / ring was clipped by `overflow` on the dialog content and form.

Centralizing the create-model dialog and aligning both forms on shared `DaDialog` / `Label` / `Input` / `Select` primitives reduces drift, and a small select-dismiss bus + singleton outside/Escape handlers fix nested modal behavior without fighting Radix event layers via DOM click hacks.

## Changes

- Added `CreateNewModelDialog` shared wrapper (`dialogTitle`, default width, `FormCreateModel`) and wired Home / Model List / Home buttons / PageHome / NewPrototypeLayout to it.
- Updated New Prototype to use `DaDialog` header title (no divider), shared field labels/inputs/selects, autofocus on Prototype Name, and required asterisk on that label.
- Updated Create Model name field autofocus and Signal label to use shared `Label`.
- Improved shared typography: larger dialog titles, field labels (`text-base`), and input/select text (`text-base`).
- Constrained `DaSelect` / `Select` dropdown width to the trigger and truncated long option text.
- Added `selectDismiss` bus and taught `Select` to close via React state; `DaDialog` dismisses open selects on outside pointerdown and Escape (dialog stays open when appropriate).
- Refactored dialog content layout so padding sits inside the scroll area (and removed form-level `overflow-y-auto`) to avoid cropping focus rings.
- Added tests for … *(none in this change; manual verification only)*.

### Before:

- Create Model dialog title/props differed by entry point; some flows omitted a proper header.
- New Prototype title lived inside the form with a different style from Create Model.
- Name fields did not consistently autofocus when the dialog opened.
- Model select menus could grow wider than the field/dialog.
- Clicking outside or pressing Escape often left the select open inside the dialog.
- Focus ring / box shadow on inputs was clipped on the top/bottom (and sometimes edges).

### After:

- All Create New Model entry points use the same titled dialog component.
- New Prototype and Create Model share dialog header and field label/input styling.
- Opening either dialog focuses the primary name input; Prototype Name shows a required `*`.
- Select menus match the trigger width; long names truncate.
- Outside click or Escape closes an open select without necessarily closing the dialog.
- Focus rings render fully around focused fields.

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing
- [ ] Existing test suite passes *(not re-run as a full suite for this PR)*

### Tested with:

- Manual check of New Prototype (`/new-prototype`) and Create New Model (home / model list / add model).
- Autofocus on name fields when dialogs open.
- Model select: width constrained; outside click and Escape dismiss the dropdown; dialog remains open where `preventOutsideClose` applies.
- Focus ring visibility on Prototype Name and other inputs after overflow fixes.
- Visual alignment of titles, labels, and inputs between Create Model and New Prototype.

## Breaking Changes

None.

## Additional Notes

- Nested Select-in-Dialog dismiss uses a lightweight module bus (`frontend/src/lib/selectDismiss.ts`) because programmatic trigger `.click()` / synthetic Escape did not close Radix Select while nested in the modal layer.
- `Select` is now lightly controlled when used uncontrolled by callers (internal open state + dismiss subscription). Call sites that already pass `open` / `onOpenChange` continue to work.
- `DaSelect` width clamping remains for any remaining legacy usages; New Prototype Model/Template fields use the shared `Select` atom.
- Possible follow-up: Playwright coverage for create-model / new-prototype dialog open → autofocus → select dismiss; consider extracting a shared “create prototype dialog” wrapper similar to `CreateNewModelDialog`.

### DIFF
--- a/frontend/src/components/atoms/DaSelect.tsx
+++ b/frontend/src/components/atoms/DaSelect.tsx
@@ -47,8 +47,12 @@ const DaSelect = React.forwardRef<
       onValueChange={onValueChange}
     >
       <div data-id={dataId} className={cn('flex flex-col', wrapperClassName)}>
-        <DaText className="font-medium!">{label}</DaText>
-        {label && <div className="pb-1"></div>}
+        {label && (
+          <>
+            <DaText className="font-medium!">{label}</DaText>
+            <div className="pb-1"></div>
+          </>
+        )}
         <SelectTrigger
           ref={ref}
           {...props}
@@ -74,7 +78,7 @@ const DaSelectItem = React.forwardRef<
   <SelectPrimitive.Item
     ref={ref}
     className={clsx(
-      'focus:bg-accent relative flex w-full cursor-pointer select-none rounded bg-white py-1.5 pl-2 pr-8 text-da-gray-medium outline-none hover:bg-gray-100 data-disabled:pointer-events-none data-disabled:opacity-50',
+      'focus:bg-accent relative flex w-full min-w-0 cursor-pointer select-none rounded bg-white py-1.5 pl-2 pr-8 text-da-gray-medium outline-none hover:bg-gray-100 data-disabled:pointer-events-none data-disabled:opacity-50',
       helperText ? 'flex' : 'items-center',
       className,
     )}
@@ -86,7 +90,9 @@ const DaSelectItem = React.forwardRef<
       </SelectPrimitive.ItemIndicator>
     </span>
 
-    <SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
+    <SelectPrimitive.ItemText className="block truncate">
+      {children}
+    </SelectPrimitive.ItemText>
     {helperText && (
       <DaText className={cn('da-label-small pl-4', helperClassName)}>
         {helperText}
@@ -128,7 +134,7 @@ const SelectContent = React.forwardRef<
     <SelectPrimitive.Content
       ref={ref}
       className={clsx(
-        'bg-popover relative z-1500 max-h-96 min-w-32 overflow-hidden rounded-md border shadow-sm data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1 data-[state=open]:animate-in data-[state=closed]:animate-out',
+        'bg-popover relative z-1500 max-h-96 w-[var(--radix-select-trigger-width)] overflow-hidden rounded-md border shadow-sm data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1 data-[state=open]:animate-in data-[state=closed]:animate-out',
         className,
       )}
       position="popper"
@@ -136,7 +142,7 @@ const SelectContent = React.forwardRef<
     >
       <SelectPrimitive.Viewport
         className={clsx(
-          'h-(--radix-select-trigger-height) w-full min-w-(--radix-select-trigger-width) bg-white p-1',
+          'h-[var(--radix-select-trigger-height)] w-full bg-white p-1',
         )}
       >
         {children}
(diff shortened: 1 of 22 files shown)


tri2510:
@7inh — review of the dialog/select UX pass. The logic is approved; one mechanical blocker (the red license check) and a scope note.

### ✅ Verified good

- **The select-dismiss architecture is the right fix** — no DOM click hacks: a small pub-sub (`selectDismiss.ts`) where each `Select` subscribes and dialogs call `dismissAllOpenSelects()`, all through React state. The controlled/uncontrolled handling is correct (uncontrolled state + `onOpenChange` kept in a ref, proper `useCallback`/effect cleanup in the subscription)
- **URL fallback is safe**: `getNavBarActionUrl` trims and falls back to `/` for empty URLs — no redirect surface, it never rewrites user-supplied URLs
- **Real consolidation**: `CreateNewModelDialog` extracted once and reused across Home / ModelList / TemplateForm entry points (net code reduction despite the new component); `FormNewPrototype` finally gets its dialog title
- **Riders are sensible**: autofocus on name and delete-confirm inputs; the staging commit is config-only (`renderPlugin` flags)
- New `selectDismiss.ts` carries its MIT header

### 🔴 Blocker: license-headers check is red on 4 files

`input.tsx`, `label.tsx`, `select.tsx`, `TemplateForm.tsx` are all missing the MIT header **on main** (pre-existing debt — same pattern as `ActionButtonsTab.tsx` before #658). This PR's touch surfaces it and the check fails. **Please add the standard header block to those four files in this PR** — clears the red check and retires the debt.

### 🟡 Scope note (non-blocking)

Five commits under a "dialog alignment" title include the nav-URL fallback and delete-dialog autofocus — related UX polish, fine as one PR, just noting the title undersells the 20-file diff.

**Verdict: approve the logic; merge-ready once the headers land.** (ECA the standing exception.)

7inh:
Added the standard Eclipse/MIT SPDX headers to:

* `frontend/src/components/atoms/input.tsx`
* `frontend/src/components/atoms/label.tsx`
* `frontend/src/components/atoms/select.tsx`
* `frontend/src/components/organisms/TemplateForm.tsx`

Confirmed locally with `scripts/check-license-headers.sh` against `main` — the check is green.

On the scope note: also agreed. This PR grew beyond dialog title/style alignment to include:

* Nav empty-URL fallback
* Delete-dialog autofocus
* Import rename UX

tri2510:
@7inh — quick flag on the latest push: `4afab1f8` (import dialog UX) is noted and consistent with the approved direction, but the **license-headers check is still red** — the four files named in the review above (`input.tsx`, `label.tsx`, `select.tsx`, `TemplateForm.tsx`) still start with imports rather than the MIT header block. The PR can't merge with that check failing; it's four copies of the standard 7-line header at the top of each file. Everything else stands approved.

tri2510:
@7inh — correction to my previous comment: **it was wrong.** The header fix (`6cfbe5fa`) landed at 06:34, three minutes *before* I posted at 06:37 flagging the check as still red — I read a stale head and credited the wrong commit. Apologies for the noise.

Verified at the actual head: all four files carry the MIT header, and the **license-headers check is green** (only ECA remains, the standing exception).

**Verdict: approve — merge-ready.**

PR #663, opened 2026-09-10
Fix vulnerabilities issues for both backend and frontend

buiduyhieu1:
Fix vulnerabilities issues for both backend and frontend


tri2510:
@buiduyhieu1 — thanks for taking on the dependency debt. I verified what I could by execution, but this PR is much larger than its title suggests and needs a few things before merge.

### ✅ Verified by execution (on this branch)

- **Frontend compiles and builds**: `tsc --noEmit` exit 0, `vite build` succeeds — React Router v7, Vite v8, eslint 9 all build clean
- **`image.js` jimp v1 migration is correct**: static `Jimp.read` ✓, named `diff` import ✓ — plus a sound robustness fix (axios `content-type` header can be non-string)
- **`customApiUtils.ts`** — same defensive handling, correct
- **Dockerfile npm-strip is legitimate scanner hygiene** — removes npm's vendored CVE-tripping packages unused at runtime, with a clear comment

### 🟠 This is a major-version overhaul, not just CVE fixes

The title undersells the breaking-change surface:

| Package | Jump | Why it needs more than a build pass |
|---|---|---|
| `multer` | 1→2 | All upload paths — including the recently hardened plugin-upload flow (#614/#619) |
| `nodemailer` | 6→10 | Skipped 4 majors; the email service (Resend integration) |
| `react-router-dom` | 6→7 | 66 files import it |
| `pm2` | 5→7 | Runtime process manager in the Docker image |
| `vite` / `eslint` / `@typescript-eslint` | 5→8 / 9.x / 8 | Toolchain (builds fine here; env variance remains) |
| `casbin` | 5.38 → **5.4.1** | ⚠️ A **downgrade** — the RBAC authorization engine going back 34 minor versions in a security PR needs an explanation |

### 🔴 Blockers / asks

1. **Explain the casbin downgrade** — if it's a typo for a newer pin, fix it; if intentional, say why in the PR body
2. **`"pm2/js-yaml": "^4.3.1"` sits in `dependencies`** — looks like the override was accidentally duplicated into deps; remove the dep entry, keep the override
3. **Runtime verification, not just build**: compile ≠ runs. At minimum a backend boot + one upload + one email-send smoke test; ideally the Playwright suites against a stack built from this branch (auth, library, admin — they exercise uploads/config paths). State what was run in the PR body.

**Verdict: request-changes** — the direction is right and the build is green, but a 9-major-bump overhaul touching uploads, email, and authorization needs the casbin question answered and runtime evidence before merge. (ECA also needs signing — first PR, welcome! Sign at https://www.eclipse.org/legal/eca/ with the email your commits use.)

buiduyhieu1:
1. casbin: ^5.4.1 was an unintended downgrade of the declared range in the vuln-fix commit; restored to ^5.38.0. Resolved version is unchanged (5.51.1) — only the lower bound was affected. Lockfile keys updated so --frozen-lockfile still passes.

2. pm2/js-yaml: not a duplicated dependency — it's in resolutions (yarn) and mirrored in overrides (npm), because the repo carries both yarn.lock and package-lock.json. The image builds with yarn, so resolutions is the load-bearing one; dropping it would un-patch js-yaml in the shipped image.

3. Verified
- Image built from this branch; Trivy HIGH/CRITICAL: 11 Node findings → 0 (2 OS-level openssl remain, unfixed upstream).
- Backend boots under pm2 v7, connects to Mongo, serves /.
- Note: npm CLI removed from the runtime image; nothing at runtime invokes it (CMD runs node via yarn; no child_process in backend/src/).

tri2510:
@buiduyhieu1 — runtime verification complete (tracked in #664), plus a thank-you for the #665/#666 follow-ups.

### ✅ Runtime E2E verification — no PR-specific regressions

Image built locally from this branch + a **main-built control image** for failure attribution; throwaway stack (isolated port/volume, guard opted in); full suite run:

| Suite | Result |
|---|---|
| auth / home-model-list / model-editable-visibility / admin | **all green** |
| home-prototype-list / prototype-extended | green on re-run / after model seeding (flake + spec precondition — both reproduce on main control) |
| **plugin-management (incl. the #614-hardened ZIP upload on multer 2)** | **2/2 ✅** |
| import-export | 1/2 — the failure **reproduces identically on main** → pre-existing (#665, root-caused in the issue) |
| nodemailer probe (resend + smtp, exercises `sendMail` directly) | clean provider/DNS errors, no module crash |
| jimp diff / sharp resize / pm2 load direct probes | correct |

**Resolved versions verified in-image:** multer 2.3.0, nodemailer 9.1.1, jimp 1.6.1, sharp 0.35.4, pm2 7.0.4, casbin 5.51.1 (unchanged from main across the range revert — results carry to `fa5e3fe9`).

### Review status after your fa5e3fe9 + explanations

- ✅ casbin downgrade — reverted, resolved version unchanged
- ✅ `pm2/js-yaml` — explained as resolutions/overrides syntax, accepted
- ✅ Runtime evidence — provided above (review ask #3)
- 🟡 Remaining nit (non-blocking): `uuid` appears in `dependencies` ^14 and `devDependencies` ^11 (frontend) — dedupe when convenient
- 🔴 **Only true blocker: ECA** — sign at https://www.eclipse.org/legal/eca/ with the email your commits use, then this is merge-ready from the review side

### Out-of-scope items (filed, nothing needed here)

- #665 — the unauthenticated model-fetch startup race (reproduces on main; separate fix branch recommended)
- #666 — E2E spec hygiene items

Issue #666, opened 2026-09-11
E2E spec hygiene: prototype-extended assumes a pre-existing model; home-prototype-list sort/category tests flake on empty listing


tri2510:
### Description

Two E2E spec problems found while runtime-verifying PR #663 against a throwaway stack (ref #664). Both reproduce **independently of that PR** (same behavior on a `main`-built stack) and make suites red or slow on fresh/clean environments.

### 1. `prototype-extended.spec.ts` requires a pre-existing model (fails on a fresh DB)

`goToFirstModelLibrary` (tests/prototype-extended.spec.ts:17-26) navigates to `/model` and waits for `a[href*="/model/"]`. On a fresh stack with zero models all 5 tests fail:

```
Error: expect(locator).toBeVisible() failed
Locator: locator('a[href*="/model/"]').first()
Expected: visible
Timeout: 8000ms  Error: element(s) not found
```

DB at the time: `models=0, prototypes=0, users=1` (admin only). After seeding one public+released model via the API, the whole suite passes (5/5, 2.0m).

**Suggested fix:** create a model in `beforeAll` (as other suites do, e.g. via `createTestModelViaApi`) instead of relying on leftover data.

### 2. `home-prototype-list.spec.ts` sort/category tests flake on a transiently-empty listing

The category tab buttons render `disabled={isEmpty}` where `isEmpty = totalResults === 0` (`frontend/src/components/organisms/HomePrototypeList.tsx:359,437`). Right after the test seeds its prototypes, the listing query can still resolve `totalResults: 0`, leaving `My Prototypes` disabled; the helper clicks it and times out:

```
Error: locator.click: Test timeout of 60000ms exceeded.
  - waiting for … getByRole('button', { name: 'My Prototypes', exact: true })
  - element is not enabled   (x87 retries)
    at helpers.ts:1366 (selectHomePrototypeCategory)
```

Observed variance on identical stacks:

| Run | Result |
|---|---|
| PR-image run 1 | 6 passed, 1 failed ("sort by Last Viewed"), 1 flaky, 2 did-not-run |
| PR-image re-run | **10 passed** |
| main-image run | 1 failed ("sort by Newest"), 1 flaky (same disabled-tab symptom), 2 passed, 6 did-not-run |

**Suggested fixes (either side):**
- Test: wait for the listing to be non-empty before clicking category/sort (the helpers already have `waitForPrototypeInHomeList`/`ensureHomePrototypeVisible`).
[...]

PR #667, opened 2026-09-11
fix(frontend): gate model/prototype detail queries on auth bootstrap (#665)

### LINKED ISSUE #665: Model detail fetch goes out unauthenticated after reload: GET /v2/models/:id 403 on private models, SPA bounces to home
### Description

`GET /v2/models/:id` is issued by the SPA **without an `Authorization` header** in some code paths (only the refresh-token cookie rides along), so an otherwise-authenticated user gets **403 Forbidden** for their own **private** model, and the SPA then navigates back to the home page.

Discovered while runtime-verifying PR #663 (ref #664): `tests/import-export.spec.ts` › "export and import prototype round-trip via library page" fails deterministically (4/4) on a throwaway production stack — **and fails identically on a stack built from `main`**, so it is pre-existing and unrelated to that PR.

### Current behavior

Playwright trace (chromium, admin logged in via UI, test created a private model + prototype, then deleted the prototype and called `page.reload()` on `/model/<id>/library/list`):

```
console error: POST /v2/auth/refresh-tokens 401 (Unauthorized)          x3
console error: GET  /v2/models/6aa3b4ce20fdc2d3eba068f9 403 (Forbidden)  x6
```

Network detail from the same trace:

- `GET /v2/models/all` → **200** — carries `Authorization: Bearer <access>` ✔
- `GET /v2/models/6aa3b4ce…` → **403** — **no `Authorization` header**, only the `Cookie: <name>=<refresh-type JWT>` ✘

The page then bounces to `/` (home), so downstream UI never renders — in the failing test, the "Import Prototype" dialog (`FormImportPrototype`, `frontend/src/pages/PagePrototypeLibrary.tsx`) never appears:

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('dialog', { name: /Import Prototype/i })
Timeout: 30000ms  Error: element(s) not found
    at helpers.ts:1711 (importPrototypeZip)
```

### Expected behavior

The model-detail fetch should go through the same authenticated client/interceptor as the list fetch (attach the in-store access token), or the auth layer should consistently restore the session after a reload; a failed model fetch should surface an error state rather than silently redirecting to home.

### Extra probe (server side is fine)

Direct curl against the same stack:

- `POST /v2/auth/refresh-tokens` with the login refresh cookie → **200** (new access token)
- `POST /v2/auth/refresh-tokens` without cookie → **401**

So the endpoint works when the cookie is sent; the defect is in which frontend path issues `GET /v2/models/:id` and how the session is (re)established after a reload.

### Suspects

- The model-detail query (e.g. `useCurrentModel` / model detail service) not using the axios instance that attaches the stored access token.
- Session restore after `page.reload()`: in-memory auth store is empty, refresh flow had already 401'd in that page lifetime, subsequent requests go out anonymous.
- Redirect-to-home on model fetch failure masking the error (observed end state: home page, no dialog, no error UI).

### Steps to reproduce

Fresh stack from image (any recent `main`), then:

```bash
cd .agents
# .env: BASE_URL/API_URL pointing at the stack, admin creds
npx playwright test tests/import-export.spec.ts -g "prototype round-trip"
```

Fails on both attempts + retries; the first test in the file (model zip round-trip) passes.

### Environment

- Production-mode Docker stack (Node 22, mongo 7.0), `NODE_ENV=production`, cookie config `secure=true, sameSite=None, httpOnly`
- Playwright chromium, workers=1, retries=1
- Reproduced on images built from `main` (`e260baa0`) and from PR #663 head

tri2510:
## What

Gate `useCurrentModel` and `useCurrentPrototype` queries on auth bootstrap, using the pattern already established in `useSelfProfile`:

```ts
enabled: !!id && (authBootstrapped || !accessToken)
```

## Why

After a page reload, `authStore` is empty until `RootLayout`'s async `refresh-tokens` restore completes. The detail hooks fired immediately (`enabled: !!id` only), so `GET /v2/models/:id` went out with **no Authorization header** → 403 on private models → the page bounced to home. The response interceptor couldn't recover because the refresh-retry requires an existing token and the failure is 403, not 401. Root-caused by @buiduyhieu1 in #665.

The `|| !accessToken` arm preserves guest behavior: when there's no token to wait for (public-viewing instances), the query fires immediately as before — only authenticated sessions wait for the restore.

## How verified

- `tsc` on the branch: only the 4 pre-existing `customApiUtils.ts` errors that exist identically on `main` (verified by stashing; none introduced)
- Guests unaffected by construction (no token → gate short-circuits open); the failing scenario (reload → private model detail) is the exact trace in #665
- Fixes #665


tri2510:
## Review — fix verified against the codebase; awaiting an independent adversarial pass before merge

**Root cause confirmed and fix direction correct:**
- `useCurrentModel.ts:22` / `useCurrentPrototype.ts:31` both gated only on `!!id` — the #665 trace (403 on private model after reload, no Authorization header, `/models/all` succeeding post-bootstrap) is exactly this race
- The `useSelfProfile.ts:23` pattern existed in-repo; adoption is consistent
- **Guest arm verified**: `enabled: !!id && (authBootstrapped || !accessToken)` — no token → fires immediately (public-viewing guests unaffected); token present → waits for bootstrap (race closed); the refresh-failure path (RootLayout's `finally` → `setAuthBootstrapped(true)` even on error) means a failed refresh also unblocks the query
- `tsc` on the branch: only the 4 pre-existing `customApiUtils.ts` errors that exist identically on `main` — none introduced

**Per the review discipline this session earned (three approvals died at the execution stage), an independent adversarial review is running** — state-machine enumeration of every `(authBootstrapped, accessToken)` combination, cache/stale-data interplay, the bootstrap-stuck-forever path, sibling hooks with the same race, and a build gate. Results land here before merge.

`Fixes #665`

tri2510:
**Correction — the first version of this fix was inert; replaced at `da4bff17`.**

An adversarial review (state-machine enumeration of every reachable `(authBootstrapped, accessToken)` store state) proved the original gate — `!!id && (authBootstrapped || !accessToken)` — was a **no-op**: the OR's `!accessToken` arm is true precisely in the bug state (post-reload: not bootstrapped, no token yet), so the query still fired unauthenticated and #665 reproduced exactly. The store's writers set the token and `authBootstrapped` atomically, so the only state the gate blocked was unreachable.

The corrected gate is the plain AND the #665 root-cause analysis actually specified (and `useSelfProfile`/`usePermissionHook` already use):

```ts
enabled: !!id && authBootstrapped
```

Safe against hangs: RootLayout's refresh runs `.finally` on success *and* failure, so bootstrap always completes — guests fire anonymously after one round-trip.

**Scope extended** per the same review: `useGetModel` and `useGetPrototype` are now gated too — `ActiveObjectManagement` is rendered by RootLayout on every page and fires the *same* `GET /models/:id` / `GET /prototypes/:id` under a different queryKey, so gating only the two `useCurrent*` hooks left the race live. Remaining id-gated hooks (list/feedback/discussion queries) are lower-severity follow-ups, tracked in #666's comment thread.

`tsc`: clean apart from the 4 pre-existing `customApiUtils.ts` errors identical on main.

PR #672, opened 2026-09-14
fix: handle template preview permissions with synthetic model IDs

7inh:
# Summary

Template previews now avoid permission checks against the synthetic `preview-template` model ID. Backend permission checks also reject non-ObjectId resource references before querying MongoDB.

It prevents template previews from sending an invalid model identifier to the permissions API and avoids treating a frontend-only preview identifier as a persisted MongoDB resource.

The preview layout uses a synthetic model ID when no real model exists. The plugin renderer did not distinguish that ID from a persisted model, so it attempted a `WRITE_MODEL` permission lookup with an invalid ObjectId.

The frontend skips an unnecessary permission request for preview-only content, while the backend now defensively rejects invalid resource IDs before database lookups.

This keeps preview behavior explicit and protects other permission callers from malformed IDs.

### Added / Improved

* Added ObjectId validation in `permission.service.js` before resource-specific permission queries.
* Added `preview-template` detection in `PluginPageRender`.
* Updated permission-hook invocation so template previews do not request model write permissions.

### Updated

* Updated preview model handling in `NewPrototypeLayout` with the synthetic `preview-template` identifier used by plugin data.
* Preserved normal permission checks for real model IDs.

### Refactored

* Separated the raw model ID from the model ID used for permission checks in `PluginPageRender`.

### Tests

* No automated tests were added in this change.

### Before

* Template previews could call `/permissions/has-permission` with `preview-template` as the model ID.
* The backend could receive a non-ObjectId value where a persisted resource ID was expected.

### After

* Template previews skip the model write-permission request when using the synthetic ID.
* Real models continue to use the existing permission flow.
* Backend permission checks return `false` for non-ObjectId resource references without querying MongoDB.

### Unit Tests

Not run; no unit tests were added for this focused change.

### Integration Tests

Not run.

### Manual Testing

Not run in a browser or running application instance.

### Existing Test Suite

Not run.

### Validation Performed

* `git diff --cached --check`

## Breaking Changes

None.

## Additional Notes

The `preview-template` value is intentionally a frontend-only sentinel and must not be treated as a persisted MongoDB ObjectId.

A follow-up could add focused frontend and backend tests covering:

* Template preview permission behavior
* Invalid resource ID handling


tri2510:
@7inh — reviewed; small, correct, defense-in-depth on both sides of the wire. **Approve.**

### Verified good

**The bug class is real** — we captured it live earlier this season: `GET /v2/permissions/has-permission?permissions=...writeModel:<invalid>` was dying as a **CastError 500** (unhandled, request hangs). This PR fixes that class at both ends:

- **Frontend**: `NewPrototypeLayout`'s synthetic `preview-template` model (confirmed at `:156`) is matched exactly; the `WRITE_MODEL` permission query is skipped entirely for it — empty `params` → the hook returns `[false]` with no request (verified against `usePermissionHook`'s `params.length > 0` enable gate). Other `model_id` consumers are unaffected: they're write flows already behind `if (!model_id)` throw-guards, and a template preview has no real model to write.
- **Backend**: the `isValidObjectId` gate in `hasPermission` turns the invalid-ref **500** into a clean `false` (fail-closed). Audited every `hasPermission` caller — all id refs are Mongo ObjectIds, user-checks pass no id, `req.body.model` is a real model id; nothing legitimate relies on non-ObjectId refs or on the 500. Bonus: this addresses the 500-vs-proper-status semantics flagged earlier as a follow-up.
- `tsc` clean (zero new errors), license check green.

### Nits (non-blocking)

- A malformed id is arguably a 400 (bad request) rather than a silent `false` — worth a thought if you touch this again, but `false` is safe and consistent with the no-id → global-check branch above it.
- `usePermissionHook(...writeModelPermission)` spread works but reads opaquely; the ternary form would be clearer.

**Verdict: approve.** (ECA the standing exception — signing at https://www.eclipse.org/legal/eca/ once covers all your PRs.)

PR #674, opened 2026-09-17
fix(home): hide hero section on small viewports

7inh:
## Summary
Hides the home page's `HomeHeroSection` banner when the viewport is both narrow and short (width < 1024px **or** height < 720px), so it no longer pushes the rest of the home page content below the fold on small screens.

## Motivation
- **What problem does it solve?** On small viewports (e.g. landscape phones, small laptops, embedded browser panes) the hero banner's fixed min/max height takes up a disproportionate share of the visible area, forcing users to scroll past it before reaching the actual page content (feature list, prototypes, models, etc.).
- **What was difficult or missing before?** There was no responsive behavior for the hero section at all — it rendered at the same height regardless of viewport size, unlike other parts of `PageHome.tsx` which already adapt to short viewports (e.g. the existing `[@media(max-height:720px)]:!pt-4` tweak).
- **Why is this approach useful?** It reuses the same Tailwind arbitrary-variant media-query pattern already established in `PageHome.tsx`, so the fix is consistent with existing conventions, requires no JS/resize listeners, and has zero runtime cost.

## Changes
- Updated `frontend/src/components/organisms/HomeHeroSection.tsx`: added a Tailwind arbitrary-variant class (`[@media(max-width:1023px)_and_(max-height:719px)]:hidden`) to the hero's outer wrapper `div`, so it's hidden only when both the width and height breakpoints are simultaneously true.

## Behavior
**Before:**
- The hero banner always renders at its configured `minHeight`/`maxHeight`, even on small/short viewports, pushing other home page sections further down.

**After:**
- The hero banner is hidden only when viewport width < 1024px **or** viewport height < 720px (both conditions must hold).
- On viewports that fail either condition (e.g. narrow-but-tall, or short-but-wide), the hero still renders as before.
- Desktop/normal viewports are unaffected.

## Testing
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing
- [ ] Existing test suite passes (not run — see Additional Notes)

Tested with:
- Manual review of the CSS change and the resulting class string.
- `npm run lint` was attempted but the environment is missing `eslint.config.js` (pre-existing, unrelated to this change), so lint could not be verified in this session.
- Visual verification via browser dev tools' responsive/device toolbar is recommended before merge:
  - Width < 1024px **or** height < 720px → hero hidden, rest of page renders normally.
  - Width < 1024px, height ≥ 720px → hero still shows.
  - Width ≥ 1024px, height < 720px → hero still shows.
  - Width ≥ 1024px, height ≥ 720px (normal desktop) → hero shows as before.

## Breaking Changes
None.

## Additional Notes
- This is a pure CSS change (Tailwind arbitrary variant), scoped directly to `HomeHeroSection` rather than conditionally wrapped in `PageHome.tsx`'s dynamic component loop, since `HomeHeroSection` is only ever used for the `'hero'` config entry type.
- Could not run `npm run lint` locally due to a missing `eslint.config.js` in this environment (pre-existing issue, unrelated to this PR) — recommend CI/reviewer confirms lint passes.


tri2510:
Reviewed — the approach is right (arbitrary-variant media query, consistent with the existing `[@media(max-height:720px)]:!pt-4` in `PageHome.tsx`, no JS, zero runtime cost). One blocking discrepancy to settle, two nits.

## 🔴 Blocker: body says `and`, code says `or`

The PR description promises: *"width < 1024px **and** height < 720px"* — hide only when both. The class implements:

```
[@media(max-width:1023px)_or_(max-height:719px)]:hidden
```

`or` hides the hero when **either** condition holds. Concrete difference:

| Viewport | `and` (as described) | `or` (as implemented) |
|---|---|---|
| Phone portrait, 390×844 | hero **visible** | hero **hidden** |
| iPad portrait, 768×1024 | hero **visible** | hero **hidden** |
| Desktop split-screen, 1920×650 | hero **visible** | hero **hidden** |
| Landscape phone, 844×390 | hidden | hidden |

With `or`, every phone and tablet loses the hero entirely, and any desktop window shorter than 720px loses it at full width — that's a significant mobile/split-screen home-page change, well beyond the summary's "narrow **and** short" framing. Notably, the motivation's own examples (small laptops — which are ≥1024px wide but short) only make sense with `or`, so I suspect the code is the intent and the description is the error — but they can't both stand.

**Please either:** change the class to `_and_` to match the description, or keep `or` and update the description to state plainly that the hero is dropped on all sub-1024px viewports and short windows.

## 🟡 Nits (non-blocking)

1. **Prefer the comma form for portability**: `[@media(max-width:1023px),(max-height:719px)]:hidden`. Comma-separated media lists are Media Queries Level 3 — universally supported; the `or` keyword is a Level 4 feature some older engines don't parse (and a parse failure silently drops the rule, so the hero would reappear on those browsers). Same semantics, zero risk.
2. **719 vs 720 seam**: the sibling tweak uses `max-height:720px` (≤720); this uses `719px` (<720). Harmless, but at exactly 720px height the two rules now disagree about what "short" means — aligning on one value would keep them consistent.

Happy to re-verify quickly once the intended semantics is confirmed — this is one class attribute away from approval.

7inh:
that was mistake when asking AI implement, recent commit to fix it

tri2510:
Thanks — the commit history (`c7012d63` `_and_` → `905ddbfa` "update chain condition to or") plus your comment settles it: **`or` is the intended semantics**, and the PR body was describing the earlier implementation. With that confirmed, the blocker is resolved. ✅

For the record, the agreed behavior: the hero is hidden on **all sub-1024px-wide viewports** (phones, tablets) **and** on any window shorter than 720px (small laptops, split-screen) — hidden only when neither applies.

Two closing notes, both optional:
- If you get a moment, a one-line touch-up of the PR description ("and" → "or") would keep it accurate for future readers — the comment thread covers it in the meantime.
- The MQ3 comma form (`[@media(max-width:1023px),(max-height:719px)]:hidden`) instead of the `or` keyword remains a nice-to-have for older-browser portability — and note it degrades safely today (browsers that don't parse `or` drop the rule and simply show the hero, i.e. current behavior).

**Approved** — merge-ready from my side. 🚀

PR #675, opened 2026-09-17
fix(prototype): hide prototype template picker when only one option

7inh:
## Summary
Hides the "Prototype Template" picker in the New Prototype form when there is only one (or zero) template option available, since there's nothing meaningful to choose between.

## Motivation
- **What problem does it solve?** The "Prototype Template *" selector was always shown whenever any project templates existed, even when there was exactly one. This forced users to interact with a dropdown that had only one possible value.
- **What was difficult or missing before?** The picker added visual clutter and an unnecessary decision point to the form for a case where no real choice exists.
- **Why is this approach useful?** The sole template was already being auto-selected via an existing effect (`FormNewPrototype.tsx`), so hiding the redundant UI doesn't change submission behavior — it just removes a no-op interaction.

## Changes
- Updated the "Prototype Template" block in `frontend/src/components/molecules/forms/FormNewPrototype.tsx` to only render the Label/Select when `templateOptions.length > 1`.
- No changes to selection/submission logic — the single template option continues to be auto-selected via the existing effect, and `selectedTemplateId` is still populated the same way.

## Behavior

**Before:**
- The "Prototype Template *" dropdown was shown any time at least one template existed, even with just one option.

**After:**
- The "Prototype Template *" dropdown is only shown when there are 2+ template options to choose from. With exactly one template, it's auto-selected silently and the field is hidden. With zero templates, nothing changes (already hidden).

## Testing
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- `npx tsc --noEmit` — typecheck passes with no errors.
- `npm run lint` could not be run in this environment (pre-existing repo issue: ESLint 9 is installed but no `eslint.config.js` exists yet — unrelated to this change).
- Manual verification recommended: run the frontend dev server and open `/new-prototype` (or the "Create New Prototype" dialog) once with a single project template available and once with multiple, confirming the picker is hidden/shown as expected and the created prototype uses the correct template in both cases.

## Breaking Changes
None.

## Additional Notes
- Scope was intentionally limited to `FormNewPrototype.tsx`. A near-identical "Prototype Template *" selector exists in `FormCreatePrototype.tsx` (used in the Prototype Library page's create dialog), but that's a separate legacy form and was explicitly excluded from this change.
- No existing convention in the codebase collapses a single-option `Select`; this introduces the pattern locally rather than reusing a shared one. A future refactor could extract this into a shared helper/component if the same need arises elsewhere.


tri2510:
Reviewed — **approved**. ✅

- The ternary refactor preserves the loading spinner and the zero-template case exactly; only the single-template case changes (picker hidden).
- I verified the linchpin: the auto-select effect (`setSelectedTemplateId(templateOptions[0].id)`) runs off `templateOptions`, independent of the Select UI — so the sole template is still selected and submission behavior is unchanged.
- No e2e or code references the `project-template-select` element expecting it always present.
- Net effect is exactly what the body promises: a no-op dropdown removed, nothing else moved.

Merge-ready once the ECA question is settled on the maintainers' side.

PR #676, opened 2026-09-17
feat(model): add HIDE_MODEL_PAGE config to hide models list page

7inh:
## Summary
Adds a new `HIDE_MODEL_PAGE` site config that lets instance admins hide the Vehicle Models list page (`/model`) without a code change, and keeps the breadcrumb trail consistent when it's hidden.

## Motivation
- **What problem does it solve?** Some instances don't want visitors browsing a global models list (e.g. single-model deployments, or instances that only want models reached through curated links). Previously there was no way to turn this page off short of editing code and routes.
- **What was difficult or missing before?** The `/model` route and its nav/breadcrumb entry were always present; hiding it required a fork-level code change instead of an instance-level config toggle.
- **Why is this approach useful?** The repo already has a backend-persisted site-config system (`useSiteConfig`) with an established convention for feature toggles like `DISABLE_CUSTOM_API_SETS`. Following that pattern means admins can flip the page on/off from the existing Site Config Management UI, consistent with how other optional features are controlled.

## Changes
- Added `HIDE_MODEL_PAGE` (boolean, default `false`) to `DEFAULT_SITE_CONFIGS` in `frontend/src/utils/siteConfig.ts`.
- Added `HIDE_MODEL_PAGE` to `PREDEFINED_SITE_CONFIGS` in `frontend/src/pages/SiteConfigManagement.tsx` so admins can manage it through the existing Site Config UI.
- Updated `frontend/src/pages/PageModelList.tsx` to read the flag via `useSiteConfig`, redirect to `/` when enabled, and short-circuit rendering (takes priority over the existing sign-in gate).
- Updated `frontend/src/components/molecules/DaBreadcrumbBar.tsx` to skip pushing the "Vehicle Models" breadcrumb entry when the flag is enabled, so child pages (e.g. `/model/:id`) show `Home > <model name>` instead of a dangling link to a hidden page.
- Updated `docs/examples/pageModels.md` to document the new config so the doc stays accurate.

## Behavior

**Before:**
- `/model` always renders the models list for any visitor (subject to the existing sign-in/`PUBLIC_VIEWING` gate).
- Breadcrumbs on model pages always show `Home > Vehicle Models > <model name>`.

**After:**
- With `HIDE_MODEL_PAGE` unset or `false`: behavior is unchanged.
- With `HIDE_MODEL_PAGE` set to `true`: visiting `/model` redirects to `/` (Home) instead of rendering the list, and breadcrumbs on model detail pages show `Home > <model name>`, omitting the now-inaccessible "Vehicle Models" crumb.

## Testing
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing
- Tested with:
  - `cd frontend && npm run build` (`tsc && vite build`) — passes with no type errors.
  - `cd frontend && npm run lint` — could not run; the checked-out `frontend/` has no `eslint.config.*` file, which ESLint 9 requires (pre-existing repo issue, unrelated to this change).
  - Manual verification of the diff/logic was done by reading the updated `PageModelList.tsx` and `DaBreadcrumbBar.tsx`; did not exercise the running app in a browser against a live backend with the flag toggled on.

## Breaking Changes
None. `HIDE_MODEL_PAGE` defaults to `false`, so existing behavior is unchanged unless an admin explicitly enables it.

## Additional Notes
- Only the `/model` list page is gated. `/model/:model_id` and its child routes (library, api, architecture, etc.) remain reachable via direct links, search, or the Home page's model cards — this was an explicit scope decision, not an oversight.
- The `/model` route itself stays registered in `frontend/src/configs/routes.tsx`; it isn't removed from the router because site config is loaded asynchronously at runtime. Hiding is enforced inside `PageModelList` via a redirect, mirroring how `DISABLE_CUSTOM_API_SETS` gates `PageVehicleApi.tsx`.
- Follow-up (not in scope here): if a future need arises to hide the entire `/model/*` tree, the same `HIDE_MODEL_PAGE` flag (or a new one) could be applied to `ModelDetailLayout` as well.


tri2510:
Reviewed — clean implementation, **approved on the code**. Two procedural items before merge.

## ✅ What checks out
- Follows the established `DISABLE_CUSTOM_API_SETS` toggle pattern exactly: `DEFAULT_SITE_CONFIGS` default + `PREDEFINED_SITE_CONFIGS` admin entry (right scope/valueType/category) — manageable from the existing Site Config UI, no new plumbing.
- The early `return null` is rules-of-hooks safe: all hooks run above it, it sits between the last `useEffect` and the pre-existing auth-gate conditional return.
- `useEffect` redirect with `{ replace: true }` + correct dep array (`[hideModelPage, navigate]`) — no history pollution, no re-trigger loops.
- Breadcrumb skip condition and dep-array update are correct; child pages degrade sensibly (`Home > <model>` without the dead `/model` link).
- Deep links to `/model/:id` still work — matches the stated intent (models reachable via curated links, only the global list is hidden).
- Docs example updated — thanks.

## 🟡 Minor notes (non-blocking)
1. **Brief flash before redirect**: `useSiteConfig` resolves async with default `false`, so the list can render momentarily before the config arrives and the redirect fires. Same trade-off as the existing config-gated features, so acceptable — just don't be surprised if a fast eye catches a frame of it.
2. **Nav link**: if an instance's configured navbar has a Models entry, clicking it now bounces to `/`. Harmless, but maybe one extra sentence in the config description ("also remove the Models nav entry if configured") would save admins a small puzzle.

## 🔴 Procedural: ECA
The ECA check fails because this branch is committed as `tranquoclinh247@gmail.com`, which isn't ECA-covered — your #674 used `til7hc@bosch.com`, which passes. Please re-push with the Bosch identity (or sign the ECA for the Gmail at eclipse.org/contribute/cla). Nothing to change in the code itself.

tri2510:
Re-reviewed at the new head `befca54c`. ✅ **Approval stands — the diff is byte-identical to the approved `c4727428`** (verified with an empty `git diff`), and the identity is now exactly right:

- Author: `7inh-bosch <til7hc@bosch.com>` ✓
- Committer: `7inh-bosch <til7hc@bosch.com>` ✓
- `Signed-off-by: 7inh-bosch <til7hc@bosch.com>` ✓

**One follow-up on the ECA check — it re-ran after your push and still fails.** Since the commit identity is now fully compliant, the failure isn't the branch's git config. Worth knowing: your merged #672 (same Bosch identity) also shows a failed ECA status, and #674 was never evaluated — so this looks like an Eclipse-account-side issue affecting your account generally, not this PR. The usual causes and fixes:

1. **ECA actually signed for `til7hc@bosch.com`?** Verify at https://www.eclipse.org/contribute/cla (sign with that exact email if not).
2. **GitHub account linked to your Eclipse account?** At https://accounts.eclipse.org → your profile → linked/social accounts, connect GitHub account `7inh`. The check's wording ("author(s) of the pull request is not covered") suggests it's resolving by the GitHub account.
3. **Re-trigger**: push any commit (an empty one is fine) or close/reopen the PR to get a fresh evaluation.

Code-wise this is done — once the ECA side clears, it's merge-ready.

tri2510:
Update from the maintainers' side: we'll proceed without the ECA check for now — no need to chase the Eclipse account linkage for this PR. 

The identity hygiene you did on the re-push (author/committer/`Signed-off-by` under `til7hc@bosch.com`) is still the right thing and appreciated — please keep doing that. The approval stands and this is merge-ready from review's perspective.

PR #677, opened 2026-09-18
fix(model): refresh sibling prototype list after deleting a model

7inh:
## Summary
Deleting a model from the home page left its (now-deleted) prototype cards visible in the sibling "All Prototypes" section until a manual page reload. This fixes the stale-cache bug by invalidating the prototype list's react-query cache when a model is deleted, and adds an e2e test that pins the fix down.

## Motivation
- **What problem does it solve?** Deleting a model also deletes all of its prototypes server-side, but the home page's prototype-list section had no way to learn that a deletion just happened, so it kept rendering cards for prototypes that no longer exist.
- **What was difficult or missing before?** `HomeModelList`'s delete handler dispatched a `window.dispatchEvent(new CustomEvent('model:list-changed'))` after deleting a model — but nothing in the codebase ever listened for that event. It was dead code masquerading as a fix.
- **Why is this approach useful?** The codebase already has an established, working pattern for this: `queryClient.invalidateQueries(...)` (~50 existing call sites), including a ready-made helper `invalidatePrototypeListQueries(queryClient)` in `usePrototypeQueries.ts` that covers every prototype list query key `HomePrototypeList` reads. Using it means the fix is a two-line change with no new plumbing, and it will keep working if more components start reading those same query keys.

## Changes
- Updated `frontend/src/components/organisms/HomeModelList.tsx`: replaced the orphaned `model:list-changed` `CustomEvent` dispatch in `handleDeleteModel` with `invalidatePrototypeListQueries(queryClient)`, using `useQueryClient()`.
- Added a `deleteModelViaHomeContextMenu` test helper in `.agents/tests/helpers.ts`, modeled on the existing `renameModelViaHomeContextMenu` helper and the `confirmNameDialog` type-to-confirm pattern already used for prototype deletion.
- Added tests for the home-page delete flow: a new e2e test `deleting a model refreshes the sibling prototype list` in `.agents/tests/home-model-list.spec.ts`, which configures both a `model-list` and `prototype-list` home section, creates a model + prototype, deletes the model via the home context menu, and asserts the prototype card disappears without a reload.

## Behavior
**Before:** Deleting a model via the home page's model card context menu removed the model card, but any prototype cards belonging to that model stayed visible in the "All Prototypes" section until the user manually reloaded the page.

**After:** Deleting a model immediately invalidates the prototype list's cached queries, so the sibling prototype-list section refetches and the deleted model's prototype cards disappear without a reload.

## Testing
- [ ] Unit tests
- [x] Integration tests (Playwright e2e)
- [ ] Manual testing

Tested with:
- `cd frontend && npm run build` (`tsc && vite build`) — typechecks cleanly.
- `cd .agents && npx playwright test tests/home-model-list.spec.ts` — full file, 17/17 tests pass (no regressions).
- Verified the new test is a real regression guard: temporarily reverted just the `HomeModelList.tsx` fix and reran the new test — it failed (stale prototype card persisted), confirming it fails without the fix and passes with it.
- `npm run lint` could not be run in this environment (pre-existing, unrelated issue: ESLint 9 is installed but the repo only ships a legacy `.eslintrc.cjs`, so `eslint` errors before reaching any file).

## Breaking Changes
None.

## Additional Notes
- The `model:list-changed` event dispatch was removed rather than kept alongside the new invalidation call, since grepping the repo confirmed it had no listeners anywhere — keeping it would have been dead code.
- No changes were needed in `HomePrototypeList.tsx`: it already reads its data via react-query, so invalidating the query keys it depends on is sufficient to trigger a refetch.
- Follow-up (out of scope here): if other pages ever need to react to model deletion beyond prototype lists, consider whether `invalidatePrototypeListQueries` should be broadened or whether model deletion should invalidate model-list queries via react-query too (currently `HomeModelList`'s own list is refreshed via a local `refetchModels()` call, not react-query).


tri2510:
Reviewed — **approved on the code**; one one-line CI fix and one procedural item.

## ✅ What checks out
- **The dead-code claim is true**: `model:list-changed` has zero listeners anywhere in `frontend/src` — the dispatch was genuinely orphaned. Good catch documenting it.
- **Invalidation coverage is correct**: `invalidatePrototypeListQueries` invalidates `prototypeQueryKeys.all` (the root key — `HomePrototypeList`'s `allForViewSort(...)` and `paged(...)` both derive from it) plus `['listModelPrototypes']`. Both sibling-list query keys go stale as needed. Dep array correctly gains `queryClient`.
- **The test is well built**: fresh API-seeded private model + prototype (tracked in `createdModelIds` for cleanup), `My Prototypes` category avoids any public-listing dependency, and the final `toHaveCount(0)` on the prototype card is precisely the regression pin. Helper mirrors `renameModelViaHomeContextMenu` + `confirmNameDialog` house style.

## 🔴 CI fix needed (one line)
The license-headers check fails: `.agents/tests/home-model-list.spec.ts` is a pre-existing headerless file surfacing on touch (same pattern as #658/#648). Per repo convention, the touching PR adds the block:

```ts
// Copyright (c) 2025 Eclipse Foundation.
//
// This program and the accompanying materials are made available under the
// terms of the MIT License which is available at
// https://opensource.org/licenses/MIT.
//
// SPDX-License-Identifier: MIT
```

## 🔴 Procedural: ECA
Same as #676 — branch committed as `tranquoclinh247@gmail.com` (not ECA-covered); your #674 used `til7hc@bosch.com`, which passes. Re-push with the Bosch identity or sign the ECA for the Gmail. No code change needed.

Happy to run the new spec against an isolated stack before merge if you want execution validation — say the word.

tri2510:
Approving and merging per maintainer decision. ✅

The code fix itself was verified in the earlier review: the dead `model:list-changed` dispatch correctly replaced by `invalidatePrototypeListQueries` (coverage over both sibling-list query keys confirmed), and the e2e test properly seeds, tracks cleanup, and pins the exact regression.

One caveat kept on record: the one-line MIT header for `.agents/tests/home-model-list.spec.ts` (CI red on the license check) is still pending — it will be handled as an immediate fix-forward on `main` so the license gate doesn't stay red for future PRs touching this spec.

PR #678, opened 2026-09-18
feat(plugin-api): add silent option to updatePrototype

7inh:
# Summary

Adds a `silent` option to the `PluginAPI.updatePrototype` method so plugins can update a prototype without triggering a success toast notification.

Plugins that call `updatePrototype` frequently or in the background — for example, autosave-like flows or silent state synchronization — were forcing a `"Prototype updated successfully"` toast on every call. This is noisy and inappropriate for updates that are not directly initiated by the user.

There was no way for a plugin to opt out of the success toast. `updatePrototype` always displayed it regardless of the calling context.

An opt-in `options.silent` flag, defaulting to `true`, keeps most calls quiet by default. Plugins that want user-visible confirmation can explicitly request it with `{ silent: false }`.

Error toasts remain unchanged, ensuring failures are still visible regardless of the `silent` setting.

### Added / Improved

* Added an optional `options?: { silent?: boolean }` second parameter to `PluginAPI.updatePrototype` in `frontend/src/types/plugin.types.ts`, with updated JSDoc.
* Added a new E2E test plugin fixture at `.agents/tests/fixtures/e2e-updateprototype-plugin/index.js`.
* The fixture exposes:

  * **Update Silent** — calls `api.updatePrototype` with the default options.
  * **Update Verbose** — calls `api.updatePrototype` with `{ silent: false }`.

### Updated

* Updated `handleUpdatePrototype` in `frontend/src/components/organisms/PluginPageRender.tsx`.
* `silent` now defaults to `true`.
* `toast.success('Prototype updated successfully')` is only triggered when `silent` is `false`.
* Error toasts for a missing `prototype_id` or an update failure remain unchanged and are always displayed.

### Refactored

None.

### Added Tests

Added `.agents/tests/plugin-api-update-prototype.spec.ts`, a Playwright spec that:

* Attaches the test plugin to a prototype tab.
* Verifies that the success toast is absent after a silent update.
* Verifies that the success toast is displayed after calling `updatePrototype` with `{ silent: false }`.

### Before

Every call to:

`api.updatePrototype(updates)`

showed a `"Prototype updated successfully"` toast, with no way for the plugin to suppress it.

### After

* `api.updatePrototype(updates)` updates the prototype silently by default.
* `api.updatePrototype(updates, { silent: true })` updates the prototype silently.
* `api.updatePrototype(updates, { silent: false })` updates the prototype and displays the success toast, matching the previous behavior.
* Error toasts are displayed in all cases and remain unchanged.

### Unit Tests

Not applicable. No unit test framework currently covers this component; the behavior is covered by the new E2E spec instead.

### Integration Tests

Not run.

### Manual Testing

Not performed in a browser during this session.

### Existing Test Suite

The frontend build passes with no new errors:

```bash
cd frontend && npm run build
```

This runs TypeScript compilation and the Vite build successfully.

`npm run lint` could not be run in this environment due to a pre-existing ESLint 9 vs. `.eslintrc` configuration mismatch unrelated to this change.

### Additional Validation

The new Playwright spec compiles and is discovered correctly:

```bash
cd .agents
npx playwright test tests/plugin-api-update-prototype.spec.ts --list
```

A full execution of the new Playwright spec against a live local stack was blocked by the repository's `e2e-env-guard`. The target backend did not have the site configuration `E2E_TEST_ENABLED=true`.

The E2E test should therefore be run in a disposable or E2E-enabled environment before merge.

## Breaking Changes

None at the API/type level. The new parameter is optional and additive, so existing single-argument calls to `updatePrototype` continue to type-check.

However, there is a **behavioral change**: the success toast is now suppressed by default. Any existing plugin that relies on the success toast appearing automatically should explicitly pass:

```ts
api.updatePrototype(updates, { silent: false })
```

## Additional Notes

* The default changes existing toast behavior for plugins that currently rely on automatic success notifications.
* The sibling `updateModel` method was intentionally left unchanged. A follow-up can add the same `silent` option if needed.
* The new E2E fixture and spec follow the existing `e2e-simple-plugin` / `plugin-management.spec.ts` conventions, including the routed external script via `page.route` and no SPDX header on spec/fixture files.


tri2510:
Reviewed — implementation is clean and the e2e test is excellent (route-fulfilled plugin script, both modes pinned, error path untouched), but **two items before merge**, plus one design point to confirm.

## 🔴 1. License headers (CI red)
Both new files lack the MIT header — `license-headers FAIL: 2 changed source file(s) missing`:
- `.agents/tests/fixtures/e2e-updateprototype-plugin/index.js`
- `.agents/tests/plugin-api-update-prototype.spec.ts`

Same block as every other source file (`// Copyright (c) 2025 Eclipse Foundation. … SPDX-License-Identifier: MIT`).

## 🔴 2. Docs not updated for a plugin-API change
`updatePrototype` gained an options parameter with a behavior-changing default, but neither `docs/guides/plugin/api-reference.md` nor the capability catalog (`docs/capabilities/plugins.md`) mentions it — and the catalog is supposed to stay code-grounded. A short paragraph + updated signature would keep them in sync.

## 🟡 3. Confirm the default: `silent = true` is a behavior change for existing plugins
`const { silent = true } = options ?? {}` means every existing plugin that calls `updatePrototype(updates)` **silently loses its success toast on upgrade** — only plugins that pass `{ silent: false }` keep it. The body states this is deliberate, and the JSDoc documents it; error toasts are unaffected. If quiet-by-default is truly the intent, fine — just flagging that it's a breaking default in observable behavior, worth a line in the docs/changelog so plugin authors aren't surprised. If you'd rather be conservative, defaulting to `false` preserves today's behavior for all existing callers and new code opts in.

Items 1–2 are mechanical; item 3 is a one-word decision. Happy to re-verify quickly after.

7inh:
updated

tri2510:
Verified at `a8a2aa09` — all three items resolved. ✅

1. **License headers** — exact MIT block on both new files; `check` is green now.
2. **Docs** — `api-reference.md` updated with the new signature and a precise description of the semantics (default `true`, `{ silent: false }` opt-out, error toasts unaffected). The capability catalog doesn't enumerate `updatePrototype`, so nothing further to sync there.
3. **Default confirmed** — documenting `default true` in the reference answers the design question: quiet-by-default stands, with an explicit opt-out for plugins that want visible confirmation.

**Approved — merge-ready.** Nice work on the e2e fixture design, by the way — route-fulfilling the plugin script keeps the test hermetic.

tri2510:
Re-verified at `63703cf6` ("update silent to false") — **the conservative default it is, and the flip is complete and consistent** across all four surfaces:

- `PluginPageRender.tsx`: `const { silent = false } = options ?? {}` — existing plugins keep the toast
- `plugin.types.ts` JSDoc: updated to match
- `api-reference.md`: updated to match
- e2e test: inverted correctly (explicit `{ silent: true }` vs no-options default), and improved — running the silent case first avoids a lingering 5s-autoClose toast tainting the `toHaveCount(0)` assertion. That ordering fix removes a latent flake. 👍

**Still approved — merge-ready.**

Note on the red license check: it is **not from this PR** — it flags `.agents/tests/home-model-list.spec.ts`, which arrived on `main` via #677 without its MIT header (known caveat from that merge). It lands in this PR's changed-file set through the merge base and will keep failing every PR until the header is added on `main`; a one-line fix-forward will clear it.

PR #673, opened 2026-09-14
chore: Update backend to support AZ KeyVault

ttntran:

PR #679, opened 2026-09-18
feat(templates): add plugin add-ons button to prototype tab bar editors

7inh:
## Summary
Adds an "Add-ons" button to the prototype tab bar editors so a plugin-backed custom tab can be attached directly from the UI — both when editing a template's Prototype Config (`/admin/templates`) and when customizing a live prototype's layout ("Customize Prototype Layout" dialog).

## Motivation
- **What problem does it solve?** The prototype tab bar list could only reorder, rename, hide, or remove tabs that were already present in the loaded config — there was no way to add a new plugin ("add-on") tab from either editor.
- **What was difficult or missing before?** Model Config tabs already had this capability via a shared `AddonSelect` plugin picker, but the picker was wired up for model tabs only. Attaching a plugin to a prototype tab required editing the underlying config outside the UI.
- **Why is this approach useful?** The add-on picker, duplicate-plugin detection, and dialog plumbing already existed and were generic enough to reuse — no new picker component or backend change was needed, just wiring the existing pieces to the prototype tab list in both places it's edited.

## Changes
- Added an "Add-ons" button to the Prototype Config → Tab Bar sub-tab in `frontend/src/components/organisms/TemplateForm.tsx`, generalizing the single shared `ModelTabAddonSelectDialog` instance to target either `modelTabs` or `prototypeTabs` via a new `addonSelectTarget` state.
- Added an "Add-ons" button to the Tab Bar tab of `frontend/src/components/organisms/CustomTabEditor.tsx` ("Customize Prototype Layout" dialog, used from the prototype detail page), reusing its existing `handleRequestAddonSelect`/`handleModelAddonSelect` handlers.
- Fixed `ModelTabAddonSelectDialog` in `CustomTabEditor.tsx` being rendered only when `mode === 'model'`, which meant the dialog never appeared for prototype ("full" mode) tab editing even though the handlers to open it already existed.
- No changes to `ModelTabListEditor.tsx`, `AddonSelect.tsx`, backend, or the `ModelTemplate`/`TabConfig` types — new tabs are appended in the same `{ type: 'custom', label, plugin }` shape already produced for model tabs.

## Behavior
**Before:**
- Prototype Config → Tab Bar (template form) and the Tab Bar tab of "Customize Prototype Layout" could only edit/reorder/hide/remove existing tabs; there was no in-UI way to add a new plugin-backed tab.
- Model Config tabs already supported adding a plugin via an "Add-ons" flow.

**After:**
- Both prototype tab bar editors show an "Add-ons" button next to the tab bar instructions. Clicking it opens the same plugin picker used for Model Config tabs, lets the user set a label, and appends the new custom tab to the prototype tab list (draggable, editable, removable, hideable like any other tab).
- Picking a plugin already used in the same tab list is blocked as a duplicate, consistent with existing Model Config behavior.

## Testing
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- `cd frontend && npx tsc --noEmit -p tsconfig.json` — no errors.
- Note: `npm run lint` could not produce a clean signal in this environment — the installed ESLint v9 is incompatible with the repo's legacy `.eslintrc.cjs` and reports ~772 pre-existing violations even on `main` before this change (confirmed by diffing baseline vs. after); this appears to be a pre-existing local tooling/environment issue unrelated to this change.
- Manual UI verification not yet performed in a running dev server (no `npm run dev` session was started for this change); recommend a maintainer or follow-up session click through both "Add-ons" buttons, add/remove/reorder a plugin tab, and confirm persistence after Save.

## Breaking Changes
None.

## Additional Notes
- The two editors (`TemplateForm.tsx` and `CustomTabEditor.tsx`) independently re-implement the prototype tab bar list UI; this PR mirrors the same fix in both rather than consolidating them into one shared component, to keep the change minimal and low-risk.
- `applyModelTabAddonSelect`, `getModelTabExcludedSlugs`, and `ModelTabAddonSelectDialog` (in `ModelTabListEditor.tsx`) are named after "model" tabs but are generic over any `TabConfig[]`; both fixes reuse them as-is without renaming, to avoid unrelated churn.
- Follow-up idea (not done here): consider extracting the duplicated prototype tab bar list markup from `TemplateForm.tsx` and `CustomTabEditor.tsx` into one shared component, since both now carry near-identical add-on wiring.


tri2510:
Reviewed — **approved**. ✅

- The reuse story holds up: in `CustomTabEditor` the select handler is already generic — `applyModelTabAddonSelect(localTabs, …)` → `setLocalTabs(result)` — and `localTabs` *is* the prototype's list when the editor runs in prototype mode, so un-gating the dialog + adding the button is correct for both modes.
- `TemplateForm` handles the dual target cleanly (`addonSelectTarget` state, unified `handleAddonSelect`, target-aware duplicate toast, dialog receives the right tab list).
- Duplicate detection and append-vs-replace (`changingPluginIndex = null`) carry over correctly.

One cosmetic nit, non-blocking: in `CustomTabEditor`, the duplicate toast still says "already added to **model** tabs" when you're in prototype mode — `TemplateForm` got the target-aware wording, this one didn't. One-line follow-up whenever convenient.

tri2510:
Heads-up @7inh-bosch — a new release is cut and waiting for your local validation before we deploy. 🚀

**`v2026.09.18`** is tagged and the image is building at `ghcr.io/eclipse-autowrx/autowrx/autowrx:v2026.09.18`. It contains your full batch plus the Key Vault work:

- #674 hide home hero on small viewports
- #675 hide single-option template picker
- #676 `HIDE_MODEL_PAGE` site config
- #677 sibling prototype-list refresh on model delete
- #678 `updatePrototype` `silent` option (default `false`)
- #679 Add-ons button on prototype tab editors
- #673 Azure Key Vault secret loading (opt-in; env mode is still the default, so local setups are unaffected)

**Not deployed anywhere yet** — the plan is for you to test locally first (image pull or `instance-setup` compose with this tag), and v3stage gets updated only after your go. Please give a shout here with anything you find.

tri2510:
Follow-up: the `v2026.09.18` build finished green ✅ — image at `ghcr.io/eclipse-autowrx/autowrx/autowrx:v2026.09.18`, and the release page has the full instance-setup bundle (compose + env sample + guide) if you'd rather spin it up with compose. Ready for your local test whenever you are.

PR #680, opened 2026-09-18
fix(model-prototype): fix admin bypass and rename addon config flag

7inh:
## Summary

Makes the Model and Prototype detail pages' "more actions" (⋮) menus behave consistently:

- the site config becomes a true **master switch** that hides the menus entirely when disabled,
- admins no longer lose the menu on models they don't own,
- menu ordering matches across both pages,
- **Save Model as Template** is now reachable from the Prototype page.

The `ALLOW_NON_ADMIN_ADDON_CONFIG` site config is renamed to `ENABLE_MODEL_CUSTOMIZATION` to match both the repo's naming convention and what the flag actually controls.

- **Admins lost the menu on models they don't own.** The Model page gate was `(isModelOwner || hasWritePermission) && flag`, with no admin bypass — unlike the Prototype page. An admin who wasn't the owner and lacked `WRITE_MODEL` saw no menu at all, contradicting the flag's own documented behavior ("Admin users can always configure addon tabs regardless of this setting").
- **Disabling the flag didn't actually disable the feature.** Because `isAdmin` bypassed the wrapper condition itself, turning the config off still left the dropdown visible to admins with a partial set of items. The flag was meant to be an off switch and wasn't one.
- **Menu order differed between the two pages**, which was inconsistent for users working across both.
- **No way to save a model as a template from the Prototype page** — a user editing a prototype had to navigate back to the Model page first.
- **The config name was wrong twice over.** `ALLOW_NON_ADMIN_ADDON_CONFIG` didn't follow the repo's `ENABLE_X` / `SHOW_X` / `DISABLE_X` convention, and `NON_ADMIN` became factually wrong once the flag started gating admin-visible actions too. It also no longer gates only "addons" — it covers layout customization and the save-as-template actions.

- The gating logic was duplicated across `ModelDetailLayout.tsx` and `PagePrototypeDetail.tsx` and had silently drifted apart, so there was no single source of truth for "can this user customize this model".
- The two kebab menus had no `data-id` hooks, so there was no stable way to assert their contents in E2E tests.

Expressing the rule as one derived boolean per page makes the precedence explicit:

```ts
flag && (isModelOwner || hasWritePermission || isAdmin)
```

The flag is absolute, and the admin bypass applies only to the ownership/permission requirement. That removes the drift between the two pages without introducing new abstractions, and it makes the config behave the way an operator would expect when they switch it off.

### Added / improved

- Added `canOpenModelMoreMenu` / `canOpenPrototypeMoreMenu` derived booleans so the site config hard-gates both menus, with the admin bypass scoped to the ownership/write-permission check only.
- Added a **Save Model as Template** item to the Prototype page's menu, reusing the existing `TemplateForm` dialog and the same `templateInitialData` construction as the Model page (`getTabConfig` / `getModelTabConfig` over `model.custom_template`).
- Added `data-id` hooks to both kebab triggers and all menu items (`btn-model-more-menu`, `btn-prototype-more-menu`, `btn-prototype-save-model-as-template`, …), matching the codebase's existing `data-id` convention, so the E2E suite has stable selectors.

### Updated

- Reordered the Model page's items so **Manage Addons** comes first and **Save Model as Template** second, mirroring the Prototype page.
- Widened the Prototype page's dropdown to `min-w-[16rem]` and added `whitespace-nowrap` to its items so the longest label ("Save Prototype as Template") renders on one line.
- Renamed `ALLOW_NON_ADMIN_ADDON_CONFIG` → `ENABLE_MODEL_CUSTOMIZATION` across:
  - the backend predefined config (`backend/src/config/predefinedSiteConfigs.js`),
  - all three frontend `useSiteConfig` reads (`ModelDetailLayout.tsx`, `PagePrototypeDetail.tsx`, `NavigationBar.tsx`),
  - the Site Config Management admin page (`SiteConfigManagement.tsx`),
  - the capability docs (`docs/capabilities/README.md`, `models.md`, `plugins.md`, `prototypes-code.md`).

  The local variable is now `enableModelCustomization`.
- Rewrote both of the config's `description` strings, which still claimed the flag was non-admin-only.

### Refactored

- Simplified the now-redundant per-item conditions inside the Prototype menu: with the wrapper guaranteeing the flag is on, **Save Model as Template** renders unconditionally and **Save Prototype as Template** reduces to `isAdmin`.

### Added tests for

- `.agents/tests/model-customization-menu.spec.ts` — 9 Playwright tests covering menu contents and ordering on both pages, the flag-off hard gate (admins included), the admin bypass on a non-owned model, the admin-only **Save Prototype as Template**, and the new Create Template dialog on the Prototype page. Captures the site-config baseline in `beforeAll` and restores it in `afterAll`.
- Registered a `test:model-customization-menu` npm script and added coverage rows to `.agents/SITEMAP.md` per the repo's E2E conventions.

### Before

| Scenario | Result |
|---|---|
| Flag **off** + admin | Model page menu still visible; Prototype page menu visible with only "Save Model as Template" |
| Flag **on** + admin who is not owner and lacks write permission | Model page menu hidden entirely |
| Model page menu order | "Save Model as Template", then "Manage Addons" |
| Prototype page menu | Only "Customize Prototype Layout" and "Save Prototype as Template" — no way to template the model |
| Site config name | `ALLOW_NON_ADMIN_ADDON_CONFIG` |

### After

| Scenario | Result |
|---|---|
| Flag **off** | Neither menu renders, for anyone, admins included |
| Flag **on** | Owners and users with write permission see the menu; admins see it even without ownership |
| Model page menu order | "Manage Addons", then "Save Model as Template" |
| Prototype page menu | "Customize Prototype Layout", "Save Model as Template", "Save Prototype as Template" (last remains admin-only) |
| Site config name | `ENABLE_MODEL_CUSTOMIZATION` |

## Testing

- [ ] Unit tests — not applicable, no unit test framework covers these components
- [ ] Integration tests — not run
- [x] Manual testing — menu width/label rendering iterated on visually; full role × flag matrix covered by E2E below

**Tested with:**

- `cd .agents && npx playwright test tests/model-customization-menu.spec.ts` — **9/9 passed** against a local stack (~1.8m). Verified afterwards that the suite restored `ENABLE_MODEL_CUSTOMIZATION` to its baseline value.
- `cd frontend && npx tsc --noEmit` — passes with no type errors.
- Repo-wide grep confirming no remaining references to the old `ALLOW_NON_ADMIN_ADDON_CONFIG` / `ENABLE_NON_ADMIN_ADDON_CONFIG` names.

> `npm run lint` could not be run in this environment — pre-existing ESLint v9 config issue (missing `eslint.config.js`), unrelated to this change.

## Breaking Changes

None for end users. Any deployment that previously overrode `ALLOW_NON_ADMIN_ADDON_CONFIG` in its database will need to re-set the same value under `ENABLE_MODEL_CUSTOMIZATION`, since this is a code-only rename with no automatic migration (confirmed acceptable — no known instance currently overrides this flag).

⚠️ **Behavior change for operators:** the flag is now absolute. Previously an admin could still reach the customization UI with the flag off; now switching it off disables the feature for everyone. That is the intended semantics, but it is a real change for any instance running with the flag disabled.

## Additional Notes

- No DB migration was added for the config rename by design (per explicit decision); if a deployment is later found to have overridden the old key, migrate its value manually via the Site Config Management page.
- The **Save Model as Template** action on the Prototype page duplicates the template-data-building logic from `ModelDetailLayout.tsx` inline rather than extracting a shared helper, to keep the change minimal; worth factoring out if a third call site appears.
- Running the E2E suite requires the target instance to declare `E2E_TEST_ENABLED=true` (the repo's fail-closed env guard); the suite mutates site config and creates/deletes test models, prototypes and users, so it belongs on a disposable stack.


tri2510:
Reviewed — **approved**. This is a thorough piece of consistency work. ✅

## What I verified
- **Precedence is now correct and identical on both pages**: `flag && (isModelOwner || hasWritePermission || isAdmin)` — the flag is a true master switch (admins included when off), and the admin bypass applies only to the ownership/permission leg. Matches the flag's original documented intent.
- **Rename is consistent everywhere**: backend `predefinedSiteConfigs.js`, frontend default + admin-UI entry, `NavigationBar`, and — appreciated — the **capability catalog updated in the same PR** (`models.md`, `plugins.md`, `prototypes-code.md`, `README`), plus SITEMAP. The new name follows the `ENABLE_X` convention and is semantically honest now that it gates layout + save-as-template, not just addons.
- **Rename migration**: body documents it as a code-only rename with no auto-migration and no known instance overriding the old key; manual re-set via Site Config Management if one surfaces. Acceptable — flagging here so it lands in the next release notes as an operator-facing change.
- **New e2e spec** (253 lines) with stable `data-id` hooks on both kebab menus + npm script — the hooks were the missing piece for asserting menu contents; good.
- New spec carries the MIT header; checks green.

One non-blocking observation: `canManageModelUI` still exists alongside `canOpenModelMoreMenu` in `ModelDetailLayout` — if the two ever drift again the bug this PR fixes could creep back. A shared hook would prevent that, but the derived-boolean-per-page approach is fine for now.

PR #681, opened 2026-09-18
fix(plugin): serialize per-slug plugin upload swap to fix ENOTEMPTY

7inh:
## Summary

Fixes a race condition in the internal plugin upload/extract flow where uploading a new ZIP for the same plugin slug could intermittently fail with a `500 Internal Server Error` caused by `ENOTEMPTY: directory not empty, rename ...`.

## Motivation

- **What problem does it solve?** When two upload requests for the same plugin slug (e.g. a double-click, a retried upload, or two admins updating the same plugin around the same time) ran close together, their directory-swap steps could interleave and corrupt the atomic "extract to temp dir → swap into place" logic, causing the second request to fail with `ENOTEMPTY` and return a 500 to the user.
- **What was difficult or missing before?** The swap logic executed three separate filesystem operations (move current plugin aside, rename new extraction into place, remove old dir) with no locking per plugin slug, and the "move current plugin aside" step swallowed *all* errors (not just the expected "doesn't exist yet" case), which hid the real failure and let execution proceed into a rename that was guaranteed to fail.
- **Why is this approach useful?** A lightweight per-slug in-process lock is the smallest change that guarantees the extract-and-swap sequence for a given slug always runs to completion before the next one starts, without introducing external locking infrastructure (e.g. Redis/DB locks) for what is a single-process filesystem operation.

## Changes

- Added a per-slug async lock (`withSlugLock`) in `backend/src/controllers/plugin.controller.js` that serializes the entire extract-and-swap sequence for a given plugin slug, so concurrent uploads to the same slug queue up and run one at a time instead of racing on the shared `pluginPath`/`oldDir` renames.
- Updated the "move current plugin aside" step to only ignore `ENOENT` (expected on a first-ever upload for a slug) and rethrow any other error, instead of silently swallowing every failure.
- No changes to the public API, request/response shape, or plugin storage layout.

## Behavior

**Before:**
- Uploading a new ZIP for a plugin while another upload for the *same slug* was in flight (or shortly after a prior request) could return `500 Internal Server Error` with `ENOTEMPTY: directory not empty, rename '.../<slug>.tmp-XXXXXX' -> '.../<slug>'`, even though each individual upload was valid.
- The underlying cause was hidden because the first rename's errors were unconditionally swallowed.

**After:**
- Concurrent/rapid uploads to the same plugin slug are processed sequentially; each one completes its extract-and-swap fully before the next begins, eliminating the `ENOTEMPTY` failure mode.
- Uploads to different slugs are unaffected and continue to run concurrently.
- Any genuine, unexpected error during the "move current plugin aside" step is now surfaced instead of being silently ignored.

## Testing

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Tested with:
- `cd backend && npx eslint src/controllers/plugin.controller.js` — no new errors (only pre-existing warnings unrelated to this change).
- `cd backend && npx jest tests/unit/controllers/plugin.upload.auth.test.js tests/unit/controllers/plugin.upload.security.test.js` — all 6/6 and existing security tests pass.
- Full `cd backend && npm test` — pre-existing, unrelated failures observed in `paginate.plugin.test.js`, `user.model.test.js`, and integration tests (environment/DB-dependent); confirmed these fail the same way without this change and are unaffected by it.

## Breaking Changes

None.

## Additional Notes

- The lock is in-process (a `Map` keyed by slug) and does not persist across server restarts or coordinate across multiple backend instances/replicas. If the plugin upload service is ever scaled horizontally, a distributed lock (e.g. via the DB or Redis) would be needed to fully eliminate the race across processes — out of scope for this fix, which targets the single-instance race that produced the reported errors.
- Considered but did not pursue: making the swap fully atomic via a single `rename` with no intermediate `oldDir` step. The existing two-step swap (move aside, then restore on failure) is kept because it protects against leaving the live plugin missing if the final rename fails; serializing per slug is sufficient to remove the race without changing that safety property.


tri2510:
Reviewed — **approved**, with one non-blocking suggestion. The race analysis and the fix are both correct. ✅

## What I verified
- **Lock mechanics are sound**: the promise chain always resolves (failures don't poison later waiters), the map entry is cleaned only when still the tail (no leak), and callers receive the *real* result/errors — only the internal chain is failure-swallowed.
- **The ENOTEMPTY mechanics are actually addressed twice over**: serialization per slug (queue instead of interleave), and the random-suffixed aside-dir (`slug.old-<hex>`) means a leftover aside-dir from a crashed upload can never block the rename. The fixed `.old` target was the structural bug.
- **Error handling got more precise, not less**: aside-rename now ignores *only* ENOENT (first upload) and rethrows everything else — previously all errors were swallowed. Swap-failure still restores the live dir.
- **`removePluginFiles` cleanup** on delete: prefix filter (`slug`, `slug.tmp-*`, `slug.old-*`) is safe against sibling-slug collisions (`foo` won't match `foo-bar`), and clearing orphans unblocks re-create under the same slug.
- **`deletePluginById` return change** (`true` → deleted doc): single caller repo-wide, and it's the one updated here. Safe.

## Non-blocking suggestion
`findEntryFile(pluginPath, …)` (and the subsequent entry-file DB write) run **outside** the lock. A second same-slug upload can execute its swap between upload A's locked section and A's entry detection — rare, but A could then ENOENT (500 despite success) or read B's entry file. Moving entry detection inside the locked section (returning the entry from it) closes the window entirely. Fine to land as-is; worth a follow-up.

Two notes for the record: the in-memory lock assumes a single backend process (true for this deployment; revisit if the backend is ever horizontally scaled), and no automated test — reasonable for a filesystem race, given the documented repro.

PR #682, opened 2026-09-18
fix(prototype): stop duplicate-name hint flash on successful create

7inh:
## Summary

Fixes a UX bug in `FormNewPrototype.tsx` where the duplicate-name warning (`DaDuplicateNameHint`) briefly flashed on screen right after a prototype was successfully created, just before the form closed or navigated away.

Submitting the "New Prototype" form with a valid, non-duplicate name would still show the "A prototype with this name already exists" hint for a split second immediately after the prototype was created — even though creation had already succeeded.

The root cause was non-obvious: on success, `handleSubmit` calls `invalidatePrototypeListQueries(queryClient)` and awaits it. That invalidation triggers a refetch of the model's prototype list, and the refetched list now includes the prototype that was just created. Since the `prototypeName` state still held that same name at that point, `useDuplicateNameCheck` momentarily (and incorrectly) flagged the name as a duplicate of itself, rendering `DaDuplicateNameHint` for one render cycle before `onSuccess`/`navigate` unmounted the form.

Clearing `prototypeName` immediately after the create call resolves — before the list-invalidation refetch completes — removes the stale state that caused the false-positive duplicate check. It's a minimal, localized fix with no changes to the duplicate-check logic itself or to the successful-creation flow.

## Changes

- Updated `handleSubmit` in `FormNewPrototype.tsx` to capture the submitted value into a local `createdPrototypeName` variable and clear the `prototypeName` state right after `createPrototypeService` resolves, before `invalidatePrototypeListQueries` runs.
- Updated the subsequent `addLog`, `toast`, and `onSuccess` calls to use `createdPrototypeName` instead of the (now-cleared) `prototypeName` state.

## Behavior

**Before:**
Submitting the form with a valid name successfully created the prototype, but the duplicate-name hint flashed beneath the input right before the dialog closed or the page navigated, making it look like something had gone wrong.

**After:**
Submitting the form with a valid name creates the prototype with no flash of the duplicate-name warning. The success toast, activity log entry, and post-create navigation/`onSuccess` callback behave exactly as before.

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested with:
- Manually created a new prototype (against an existing model, and while creating a new model at the same time) and confirmed the duplicate-name hint no longer flashes on successful submit.
- Confirmed the toast message, activity log entry, and post-create navigation/`onSuccess` behavior are unchanged.

## Breaking Changes

None.

## Additional Notes

- This is a narrow, targeted fix — it does not touch `useDuplicateNameCheck`, `DaDuplicateNameHint`, or the model-name duplicate check path, since the flash was specific to the prototype-name check racing against the post-create query invalidation.
- No automated test was added for this timing-dependent UI flash; covering it would require simulating the query invalidation/refetch race in a component test. Flagging as a possible follow-up if regression coverage is desired.


tri2510:
Reviewed — **approved**. Minimal and precisely targeted. ✅

The root cause is exactly as described: `invalidatePrototypeListQueries` → refetch now contains the just-created prototype → `useDuplicateNameCheck` momentarily flags the still-set `prototypeName` as a duplicate of itself. The fix snapshots the name (`createdPrototypeName`), clears the state **before** the invalidation refetch can land, and moves every downstream consumer (log, toast, `onSuccess`) to the snapshot — so nothing else changes behavior. Correct ordering, correct scope.

PR #683, opened 2026-09-18
fix(runtime-preview): unblock preview iframe subresources

7inh:
## Summary

The Runtime Preview widget renders the running prototype's UI inside an iframe served through the `/runtime-preview/<runtime-id>/` proxy. Since the iframe sandbox was introduced, that frame has been unable to load any of its own subresources, so the preview never renders — it only shows NiceGUI's *"Connection lost. Trying to reconnect..."* overlay. This PR restores subresource loading and clears the follow-on blocker for the runtime's websocket connection.

## Motivation

**What problem does it solve?**

`8826ead9 fix(runtime-preview): add iframe sandbox and auth gate on proxy route` added `sandbox="allow-scripts allow-forms"` to the preview iframe. Because `allow-same-origin` was omitted, the framed document is placed in an **opaque origin**. Every subresource it then requests from the application host is treated as cross-origin, and the app's `Cross-Origin-Resource-Policy: same-origin` header (helmet default) rejects them.

Observed in DEV DevTools:

- `nicegui.css`, `nicegui.js`, `socket.io.min.js`, `vue.global.prod.js`, `quasar.umd.prod.js` → `307` → redirected request fails with `net::ERR_BLOCKED_BY_RESPONSE`
- `query.js`, `echarts.min.js`, `echart.js`, `timer.js` → `CORS error`

With the NiceGUI client bundle never loading, socket.io is never established, which surfaces to the user as the "Connection lost" overlay.

**What was difficult or missing before?**

The failure was environment-dependent in a way that masked the real cause:

| Environment | Behavior | Reason |
|---|---|---|
| QA (tag `3.5.1`) | Works | Predates `8826ead9` — no sandbox, frame is same-origin |
| DEV (current) | Broken | Has the sandbox → opaque origin → CORP blocks everything |
| Local (current) | Appeared to work | The preview was only ever exercised via the runtime's direct host port, bypassing the iframe entirely |

Because QA and local both appeared healthy, this looked like a DEV infrastructure problem (ingress, TLS, auth gateway, websocket proxying) rather than an application regression. The auth gate portion of `8826ead9` is correct and unaffected; only the sandbox attribute is at fault.

**Why is this approach useful?**

Adding `allow-same-origin` restores the origin the frame is actually served from — the preview is proxied through the application's own origin — so CORP, CORS, cookies, and socket.io all behave normally again. The remaining sandbox flags continue to block top-level navigation, popups, pointer lock and plugins, so the frame stays more constrained than it was before `8826ead9`, which shipped with no sandbox at all.

## Changes

- **Updated** `backend/static/builtin-widgets/runtime-preview/index.html` — added `allow-same-origin` to the preview iframe's `sandbox` attribute, with an inline comment recording why it is required, so the flag is not removed again as a hardening change.
- **Updated** `backend/src/app.js` — added `ws:` and `wss:` to `connectSrc` in the **production** CSP block, mirroring the development block which already allows them. Prevents the runtime's socket.io connection from being blocked by CSP once the bundle loads.
- No refactoring, no new tests (see *Testing*).

## Behavior

**Before:**

- Opening a prototype's Runtime Preview widget shows an empty frame with *"Connection lost. Trying to reconnect..."*.
- The NiceGUI asset requests fail with `net::ERR_BLOCKED_BY_RESPONSE` / `CORS error`.
- No `socket.io` / `_nicegui_ws` connection is ever attempted, because the client bundle never initialises.
- Affects every deployment running a build that includes `8826ead9`; older builds (e.g. QA on `3.5.1`) are unaffected.

**After:**

- The preview frame loads the NiceGUI bundle normally and renders the running prototype's UI.
- The runtime's socket.io connection is permitted by CSP in production as well as development.
- The auth gate on the proxy route and the remaining sandbox restrictions are unchanged.

## Testing

- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

**Tested with:**

- Root cause confirmed by inspecting DEV (`*.app.cloudhub.etas.com`) DevTools Network output — the `307` → `ERR_BLOCKED_BY_RESPONSE` pattern on NiceGUI assets and `CORS error` on the chart scripts, all initiated by the `/runtime-preview/PRIVATE-04-…` document.
- Regression origin confirmed via `git log -p --follow` on `backend/static/builtin-widgets/runtime-preview/index.html`, which shows the `sandbox` attribute first appearing in `8826ead9`.
- Cross-checked against QA (tag `3.5.1`), which predates that commit and renders the preview correctly.

> **Not yet verified end-to-end.** The changed file is baked into the image, so the fix cannot be exercised until a new image is built and deployed. Manual verification in DEV is pending: after deploy, the `nicegui.*` requests should return `200` instead of `ERR_BLOCKED_BY_RESPONSE`, and the `socket.io` requests (which begin as XHR long-polling, not WebSocket) should succeed. The existing test suite has not been run against this change.

## Breaking Changes

None.

## Additional Notes

**Security trade-off — please review deliberately.** `allow-scripts` combined with `allow-same-origin` effectively disables the sandbox's origin isolation: the framed content — which is *user-authored prototype code* running on a runtime kit — regains the application's origin and can reach the parent page's cookies, storage and DOM. This is the risk `8826ead9` was reaching for. Two mitigating points:

1. The preview is proxied through the application's own origin regardless, and the pre-`8826ead9` behaviour (still shipping in QA) had no sandbox at all — so this restores the previous posture rather than weakening it further.
2. The other sandbox flags remain in force, so this is still stricter than QA.

**Stricter alternative, if the isolation is considered worth keeping.** Retain the sandbox as-is and instead serve `/runtime-preview/*` with `Cross-Origin-Resource-Policy: cross-origin` — set on both the route response and `on.proxyRes` in `app.js`. That preserves the opaque origin, but socket.io's polling transport would then issue requests with `Origin: null`, which the CORS configuration would have to accept; `Access-Control-Allow-Origin: null` combined with credentials is not permitted, so this path needs design work and testing before it can be relied on. Reasonable follow-up if the security posture matters more than the one-line fix.

**Downstream copy must be synced.** The customer composition repository ships its own overlay copy of this file at `image-content/playground-be/static/builtin-widgets/runtime-preview/index.html`, which is baked *on top of* the base image. That copy carries the identical `sandbox` line and will override this fix unless the same change is applied there and the pinned image digest is updated. Deployments will not pick up this fix until that happens.

**Related latent bug (not fixed here).** The composition's `nginx/generate-nginx-config.sh` normalises the preview URL with `return 301 /runtime-preview/${runtime_id}/;`. nginx expands that to an absolute URL built from `$host`, which drops both the non-standard port and the query string — so an `access_token` on the no-trailing-slash form is silently stripped. It does not affect DEV today (that URL currently carries no token, and DEV is on port 443), but it does affect local setups on custom ports and will bite as soon as a token is present. Fix is `return 301 $scheme://$http_host/runtime-preview/${runtime_id}/$is_args$args;`, tracked separately in the deployment repository.


tri2510:
Reviewed — **approved, with one security note stated plainly** (for the record, not as a blocker). The diagnosis is excellent and the fix is the right pragmatic call. ✅

## What checks out
- **Root cause confirmed**: `sandbox` without `allow-same-origin` ⇒ opaque origin ⇒ helmet's CORP `same-origin` rejects every subresource; the NiceUI bundle never loads, socket.io never connects, "Connection lost" overlay. The environment matrix (QA predates the sandbox, DEV has it, local bypassed the iframe entirely) explains exactly why this looked like infra.
- **The CSP half is correct and minimal**: `*` in `connect-src` does not cover the `ws:`/`wss:` schemes — they must be listed explicitly, and the production block now mirrors the dev block that already had them.
- **The auth gate from `8826ead9` is untouched** — the proxy still strips `authorization` and filters cookies toward the runtime.
- Inline comment on the iframe explaining *why* `allow-same-origin` is required — good; that's what stops the next hardening pass from blindly removing it.

## The security note (worth having in writing)
`sandbox="allow-scripts allow-same-origin"` on a frame served **from the app's own origin** is, per the spec's own warning, self-strippable — the framed document can reach `window.parent` and remove its sandbox. So the remaining flags (no popups, no top-navigation…) are advisory, and the body's "stays more constrained than before `8826ead9`" overstates the guarantee. That said, this is acceptable in context because:

1. It restores the exact status quo QA ships today (`8826ead9`'s sandbox never actually worked anywhere it applied);
2. Prototype-authored JS already runs same-origin in dashboard widgets, so this iframe was never the outer boundary for prototype code anyway;
3. Tokens aren't exposed in `localStorage`, and the proxy gate limits what flows *toward* the runtime.

**Recommended follow-up (separate issue)**: serve the preview frame from a distinct origin (separate subdomain/port for the runtime-preview proxy). That's the only configuration in which `allow-same-origin` + `allow-scripts` is genuinely safe and the sandbox becomes a real boundary. Happy to help spec it.

